PR #28, opened 2026-01-15
feat(sandbox): add convience kwargs to stream output without manual iteration

domphan-wandb:
### Summary

Adds a `print_output` parameter to `sandbox.exec()` that controls whether output is printed in real-time. By default, `exec()` is silent (output available via `result.stdout`). For convenience during debugging, set `print_output=True` to auto-print output without manual iteration.

### Motivation

There are use cases where streaming output is needed—watching long-running processes, debugging, or monitoring logs. Previously this required manual iteration:

```python
process = sandbox.exec(["python", "script.py"])
for line in process.stdout:
    print(line, end="")
result = process.result()
```

This PR adds a convenience method to avoid that boilerplate:

```python
result = sandbox.exec(["python", "script.py"], print_output=True).result()
```

This pattern will also be useful for a future CLI where `aviato exec` can leverage the same printing behavior.

### API

- `print_output: bool = False` — new parameter on `exec()`
- `AVIATO_EXEC_PRINT=1` — environment variable to enable printing globally

| `print_output` | Behavior |
|----------------|----------|
| `False` (default) | Silent — access output via `result.stdout` |
| `True` | Auto-print — output printed in real-time |

### Design Decisions

**Opt-in printing (silent by default)**

We chose to make printing opt-in rather than opt-out because:
- SDK users typically want programmatic control over output
- Unexpected printing can pollute logs in automated pipelines
- Explicit `print_output=True` clearly signals intent

**No prefix on output**

Currently, auto-printed output has no prefix (e.g., sandbox ID). This keeps the convenience option simple and unopinionated. If prefixed output is needed (e.g., for multi-sandbox scenarios), users can use manual iteration with their own formatting. We can revisit adding an optional prefix in the future if there's demand.

### How to Review

1. **Implementation**: Check `src/aviato/_sandbox.py` — the `print_output` parameter flows to `_exec_streaming_async()` where printing happens
2. **Environment variable logic**: Verify `AVIATO_EXEC_PRINT` correctly overrides `print_output=False`
3. **Tests**: Unit tests cover silent default, `print_output=True`, stderr handling, and env var behavior
4. **Run the example**: `python examples/streaming_exec.py` demonstrates all three output patterns

### Testing

```bash
# Unit tests
uv run pytest tests/unit/aviato/test_sandbox.py -v -k "print_output or silent or env_var"

# Integration tests
uv run pytest tests/integration/aviato/test_sandbox.py -v -k "print_output"
```


iiilisan (dismissed):

NavarrePratt (commented):
Few small comments, but lgtm after cleaning them up.

NavarrePratt (approved):

domphan-wandb:
stale

PR #84, opened 2026-03-13
chore(proto): vendor secrets_pb2_grpc and harden update script

NavarrePratt:
Add missing secrets_pb2_grpc.py to the vendored proto directory and the PROTO_FILES list in update-protos.sh. Also skip missing .pyi files in post-processing so local generation (which lacks .pyi) does not fail.


NavarrePratt:
Closing - the protobuf version check is stale after recent proto bumps, and secrets_pb2_grpc.py has no consumers. The file-existence guards are a minor edge case we can fix inline if needed.

PR #85, opened 2026-03-20
feat(examples): add nanochat training pipeline example

juan-lee:
Adds an end-to-end LLM training example using karpathy/nanochat inside a single sandbox, demonstrating long-running sequential pipelines with per-step exec() streaming.


NavarrePratt:
@juan-lee Do we want to get this PR over the line?

PR #90, opened 2026-03-23
fix: enforce keyword args on NetworkOptions and Secret

nicholaspun-wandb:
Split from: https://github.com/coreweave/cwsandbox-client/pull/78

Keep the API unambiguous by enforcing keyword arguments, making positional construction not possible. This is already enforced at the type checking level so no changes needed in our code, tests or examples. For users of the SDK, this prevents them from accidentally defining the wrong field when there are multiple optional fields. For example, with `Secret`, the `field` and `env_var` fields are both optional, with `field` coming before `env_var`. However, a user wanting to assign an environment variable without using the `field` field could accidentally write `Secret("wandb", "foo", "FOO")` which will assign `"FOO"` to `field`

### DIFF
--- a/src/cwsandbox/_types.py
+++ b/src/cwsandbox/_types.py
@@ -123,7 +123,7 @@ class ExecOutcome(StrEnum):
     FAILURE = "failure"
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, kw_only=True)
 class NetworkOptions:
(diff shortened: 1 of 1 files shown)


abatilo (dismissed):
I'm a little worried about this with it technically being a breaking change for NetworkOptions but I guess we're not actually GA yet.

abatilo (approved):

PR #98, opened 2026-04-03
feat: rename aviato → sandbox (align with server rebrand)

abatilo:
Update vendored proto stubs and SDK to match the server-side rename
from aviato/v1beta1 to sandbox/v1beta2:

- Proto: atc_pb2 → gateway_pb2, ATCService → GatewayService,
  ATCStreamingService → GatewayStreamingService
- Public API: tower_ids → runner_ids, runway_ids → profile_ids,
  tower_group_id → runner_group_id
- CLI flags: --tower-id → --runner-id, --runway-id → --profile-id
- All gRPC service paths updated to coreweave.sandbox.v1beta2.*

Generated stubs from aviato repo's `make buf-gen-python` output.
All 658 unit tests pass.

PR #108, opened 2026-04-11
feat: add TERMINATING status and _Stopping lifecycle state

NavarrePratt:
## Summary

The backend added `SANDBOX_STATUS_TERMINATING = 9` as a non-terminal intermediate state for sandboxes being torn down. Previously the SDK received proto value 9 and silently fell back to `UNSPECIFIED`. This PR adds proper handling with a new `_Stopping` lifecycle state that preserves metadata and lets the SDK poll through the drain to the real terminal state (COMPLETED or FAILED).

## Design decisions

**_Stopping is its own lifecycle state, not _Starting or _Terminal.** _Starting drops metadata (runner_id, profile_id, started_at). _Terminal would end polling prematurely. _Stopping preserves metadata and allows the complete-poll task to continue to the real terminal state.

**stop() sends the RPC first, then transitions to _Stopping.** No rollback needed on RPC failure since state has not changed yet. On success, `_running_task` is cancelled (sandbox will not become RUNNING) but `_complete_task` stays alive to poll to terminal.

**Shared _stop_task pattern.** First `stop()` call creates the task, subsequent callers join via `asyncio.shield`. First caller parameters (snapshot_on_stop, graceful_shutdown_seconds) win.

**_do_stop always polls to terminal.** Even when `_Stopping` was set by a background poll before `stop()` is called, `_do_stop` skips the RPC but falls through to `_await_terminal_after_stop()` to ensure `stop().result()` resolves at terminal state. `_stop_owned` stays False in this path (correct provenance).

**raise_on_termination uses `_stop_owned`, not `_termination_observed`.** An earlier version used `_termination_observed` (set whenever TERMINATING was seen) but that caused false-positive raises on normal exits that polled through TERMINATING. `_stop_owned` is local provenance: "this client sent a successful Stop RPC." It will not fire for normal exits. Known limitation: external kills (infrastructure, lifetime, other clients) that result in COMPLETED will not raise because `_stop_owned` is False. Requires backend termination_reason metadata to fully close this gap.

**_do_poll_running logs instead of raising for _Stopping.** A short-lived sandbox can go RUNNING -> TERMINATING -> COMPLETED during startup. Now it logs and returns, letting the caller handle the terminal state normally.

**State regression guard.** Once in _Stopping, `_apply_sandbox_info` only allows forward transitions to _Terminal. Stale poll responses reporting RUNNING are rejected.

**stream_logs(follow=False) allowed during _Stopping** for historical log retrieval. `follow=True` is blocked since the sandbox is draining.

## What changed

- Bump vendored proto stubs from BSR (`20260409183835+605a52578aa2`)
- `SandboxStatus.TERMINATING` added to StrEnum; `TERMINATED` deprecated in docstring
- `_Stopping` lifecycle state with metadata fields and state regression guard
- Shared `_stop_task` pattern with `_await_terminal_after_stop()` helper
- `_stop_owned` provenance for `raise_on_termination`
- Operation guards block exec/shell/read/write during `_Stopping`
- `get_status()` non-caching for `_Stopping` (mutable state)
- Fix pre-existing `KeyError` bug in `from_proto()` exception handler
- AGENTS.md updated with new lifecycle docs

## Test plan

- [x] `from_proto(9)` returns TERMINATING
- [x] Unknown proto name triggers KeyError path, falls back to UNSPECIFIED
- [x] `_lifecycle_state_from_info(TERMINATING)` returns `_Stopping` with metadata
- [x] State transitions: `_Stopping` -> `_Terminal` allowed, `_Stopping` -> `_Running` rejected
- [x] Operations blocked during `_Stopping` (exec, shell, read_file, write_file)
- [x] `stream_logs(follow=True)` blocked, `follow=False` allowed during `_Stopping`
- [x] `_do_poll_running` logs and returns for `_Stopping` (no raise)
- [x] `wait_until_complete()` polls through TERMINATING to terminal
- [x] Normal exit through TERMINATING does NOT raise with raise_on_termination=True
- [x] `stop()` + terminal COMPLETED raises via `_stop_owned`
- [x] `stop()` on externally-set `_Stopping` polls to terminal (not early return)
- [x] `stop()` RPC failure preserves `_Running` state
- [x] Repeated `stop()` joins shared task
- [x] Session.close() with `_Stopping` sandbox joins stop task
- [x] Property accessors work during `_Stopping`
- [x] CancelledError during `_Stopping` -> `SandboxNotRunningError`
- [x] `Sandbox.from_id()` with TERMINATING sandbox
- [x] `get_status()` not cached during `_Stopping`, cached during `_Terminal`
- [x] Integration: `test_stop_reaches_terminal_through_terminating`
- [x] Integration: `test_stop_raise_on_termination_with_stop_owned`
- [x] Integration: `test_natural_exit_no_raise_on_termination`
- [x] `mise run test` passes
- [x] `mise run typecheck` passes
- [x] `mise run lint:check` passes


NavarrePratt:
This change is part of the following stack:

- #98
    - #108 ◀
        - #109

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

abatilo (approved):

PR #123, opened 2026-05-17
feat: add Harbor environment adapter as optional extra

abatilo:
## Summary

Adds `CWSandboxEnvironment` as a [Harbor](https://github.com/harbor-framework/harbor) `BaseEnvironment` subclass, installable via a new `harbor` optional extra. The adapter was previously vendored inside a downstream Harbor deployment — moving it next to the SDK lets updates ship on the cwsandbox release cadence and removes a piece of duplicated maintenance surface.

```bash
pip install "harbor>=0.6.6" "cwsandbox[harbor]"

harbor run ... \
  --environment-import-path cwsandbox.harbor:CWSandboxEnvironment \
  --ek docker_image=<your-pre-built-image>
```

Harbor's existing `--environment-import-path` mechanism resolves this via `importlib`; no Harbor-side changes are required.

## What's in this PR

- `src/cwsandbox/harbor.py` — the adapter (`CWSandboxEnvironment`), fully strict-mode mypy clean.
- `pyproject.toml` — new `harbor` optional extra, new `[tool.uv] conflicts` block, scoped mypy override for `harbor.*`, `tenacity` added to `dev` for type-stub visibility.
- `tests/unit/cwsandbox/test_harbor_adapter.py` — `importorskip`-gated smoke test asserting `CWSandboxEnvironment.type() == "cwsandbox"`.
- `README.md` — new "Using with Harbor" section.
- `DEVELOPMENT.md` — new "Iterating on the Harbor adapter" subsection documenting the separate-venv dev loop.

## Notable packaging decisions (worth a reviewer's attention)

1. **Python version gate.** The `harbor` dep has a `python_version >= '3.12'` marker because Harbor requires Python 3.12+ while cwsandbox supports 3.11+. On 3.11 the dep silently won't install — matching cwsandbox's wider support floor.

2. **`[tool.uv] conflicts` declaring `harbor` mutually exclusive with `cli` and `dev`.** Harbor's transitive deps (LiteLLM specifically) pin `click<8.2`, while both `cli` and `dev` require `click>=8.2` for the CLI's stdout/stderr stream split (introduced in click 8.2.0, see [#89](https://github.com/coreweave/cwsandbox-client/pull/89)). The conflict is real, and the audiences (CLI binary users vs. Harbor framework users) are disjoint in practice. This means adapter contributors use a separate `.venv-harbor` venv — documented in DEVELOPMENT.md.

3. **Tenacity added to `dev`.** Tenacity is a runtime dep of the adapter (retry decorators) and is in the `harbor` extra accordingly. It's also added to `dev` so its bundled type stubs are visible to mypy in the default dev venv — without this, every `@retry` flags `untyped-decorator`.

4. **Scoped mypy override for `harbor.*`.** Since the harbor package can't coexist with the dev venv (per the conflict above), mypy treats `harbor.*` modules as `Any` via `ignore_missing_imports + follow_imports = skip`. The body of `cwsandbox.harbor` is still fully strict-checked.

5. **`__init__.py` is NOT touched.** The adapter is intentionally accessed only as `cwsandbox.harbor:CWSandboxEnvironment`, not re-exported from `cwsandbox/__init__.py`, so `import cwsandbox` doesn't accidentally pull harbor into minimal installs.

## Verification

- `uv sync --extra dev`: passes.
- `uv sync --extra harbor` (in isolation): passes.
- `uv run pytest tests/unit/`: 1091 passed, 1 skipped (the smoke test is skipped when harbor isn't installed in the dev venv).
- `uv run mypy src/`: clean across all 31 source files.
- `uv run ruff format --check .` and `uv run ruff check .`: clean.
- End-to-end on a Python 3.12 scratch venv with `pip install -e '.[harbor]'`: `CWSandboxEnvironment.type()` returns `cwsandbox`.

## Test plan

- [ ] CI: existing pytest/mypy/ruff matrix passes
- [ ] Reviewer: verify the `[tool.uv] conflicts` semantics make sense for this repo's audience split
- [ ] Reviewer: confirm `python_version >= '3.12'` gate is the right shape for the dep matrix
- [ ] Post-merge: confirm semantic-release picks up the `feat:` subject and publishes `0.24.0` to PyPI
- [ ] Post-publish: `pip install "harbor>=0.6.6" "cwsandbox[harbor]==0.24.0"` and import-smoke from a clean Python 3.12 venv


NavarrePratt:
An official integration is now upstreamed into harbor so this isn't needed.

abatilo:
Hooray

PR #126, opened 2026-05-20
fix: reject bare string sandbox tags

nicholaspun-wandb:
JIRA: https://coreweave.atlassian.net/browse/WB-33933

## Description

Resolves https://github.com/coreweave/aviato/issues/984. Normalize and validate tags so that single string and non-string tags are rejected. 

## Testing

Updated and added unit tests

### DIFF
--- a/src/cwsandbox/_defaults.py
+++ b/src/cwsandbox/_defaults.py
@@ -111,6 +111,24 @@ def _merge_dicts(base: dict[str, str], additional: dict[str, str] | None) -> dic
     return merged
 
 
+def _normalize_tags(tags: Iterable[str] | None) -> tuple[str, ...]:
+    """Normalize tag sequences and reject bare strings.
+
+    Strings are iterable in Python, so accepting one would split it into
+    characters (``"prod"`` -> ``("p", "r", "o", "d")``).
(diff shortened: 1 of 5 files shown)


src/cwsandbox/_defaults.py
```diff
@@ -231,16 +246,17 @@ def __post_init__(self) -> None:
...
+    def merge_tags(self, additional: Iterable[str] | None) -> list[str]:
         """Combine default tags with additional tags.
 
         Tags from both sources are included. Order is: defaults first,
         then additional tags appended.
         """
-        base = list(self.tags)
-        if additional:
-            base.extend(additional)
+        base = list(_normalize_tags(self.tags))
```
NavarrePratt: nit: we don't need to call `_normalize_tags` again, the `__post_init__` already takes care of it.

NavarrePratt (dismissed):
one small nit, but otherwise looks good!

src/cwsandbox/_defaults.py
nicholaspun-wandb: removed!

NavarrePratt (approved):

PR #130, opened 2026-05-26
feat(logs): auto-resume stream_logs across transient disconnects

brandonrjacobs:
## Summary
- `stream_logs(follow=True)` now captures the per-frame session id and cumulative byte offset and, on a transient transport error, reconnects with those fields so the live tail continues without the caller seeing the disconnect.
- Servers that don't speak the resume wire protocol return empty values for the new fields; the resume branch is never entered and behavior is unchanged.
- In-band `SESSION_NOT_FOUND` / `INVALID_RESUME_OFFSET` rejects fall back to a fresh init; `REPLAY_GAP` is logged and the stream continues.
- Regenerates the vendored streaming proto stubs to include the new fields. Pinned to protobuf 5.26.1 per `scripts/update-protos.sh` policy.

## Test plan
- [x] `mise run lint` / `ruff format --check .` clean
- [x] `mypy src/` clean
- [x] `mise run test` — 1117 unit tests pass (8 new)
- [x] `mise run test:e2e` against staging — 113 passed, 2 skipped (env-gated); 3 new resume tests pass


abatilo (dismissed):

NavarrePratt (commented):
[via Claude]

# Branch review — feat/log-stream-resume

Multi-agent review against this branch and the wire contract in `streaming_pb2.pyi`. The most recent commit (the wire-contract fix) is a nicely surgical landing — it picked up the REPLAY_GAP / INVALID_RESUME_OFFSET / RUNNER_* handling, the partial-line buffer corruption on fresh fallback, the lost `__cause__` on resume exhaustion, the dropped `attempt+1 < MAX` gate, and the deterministic state-machine tests. The new `_STREAMING_FRESH_REINIT_CODES` set is exactly the table-driven dispatcher this code wanted.

**Five things still open** — see inline comments:

- Fresh-init reconnects still re-emit `tail_lines` / `since_time` on every attempt, despite the in-code comment promising "no replay of older bytes."
- A transient transport error *before* the first frame arrives is never retried — the retry guard requires a `session_id` we don't have yet.
- Exceptions raised inside the retry block can be silently swallowed: the outer `finally` enqueues the EOF sentinel before the outer `except` enqueues the exception.
- `stub` is acquired once outside the retry loop and reused across attempts; works today, but couples retry correctness to gRPC subchannel behavior and to the channel never being torn down externally.
- The streaming retry classifier dispatches on raw `StatusCode` only; AIP-193 reason metadata that `_translate_rpc_error` would consult is invisible to it. Forward-compat note, not a live bug.

**One non-blocking structural note for when exec-resume gets wired:**

The proto stubs already ship `ExecStreamInit.resume_session_id` and `StreamingExecReady.session_id`. When exec resume gets added, three near-identical retry loops (log + exec-TTY + exec-non-TTY) will exist and drift independently. Worth introducing a small generic primitive now — most of the pieces are here:

- `_LogStreamResumeState` dataclass for the cross-attempt state (today scattered as bare locals: `session_id` / `last_offset` / `line_parts` / `line_parts_bytes` / `last_transport_error`), with a `clear_resume()` that pairs the session+offset+buffer reset together so the invariant is type-enforced.
- `attempt_outcome` collapses to `should_retry: bool` (the fresh-vs-resume distinction is already expressed by inline mutation of `session_id`).
- A generic `_run_resumable_stream(open_attempt, frame_handler, classifier, defaults)` that takes the log/exec-specific frame handlers and in-band classifiers as inputs, so exec resume is additive instead of a copy of the entire control flow.

src/cwsandbox/_sandbox.py
```diff
@@ -2914,130 +2993,309 @@ async def _stream_logs_async(
...
+                                # Per the wire contract: when resume_session_id is
+                                # set, all fields except sandbox_id are ignored.
+                                # Keep the others off the wire to make that
+                                # explicit and avoid log noise from servers that
+                                # validate the rule.
+                                init.resume_session_id = resume_id
+                                init.resume_offset = resume_off
+                            else:
+                                if tail_lines is not None:
+                                    init.tail_lines = tail_lines
```
NavarrePratt, line 3107: [via Claude] **High: Fresh-init reconnects (SESSION_NOT_FOUND, REPLAY_GAP, RUNNER_*) re-emit `tail_lines` / `since_time` on every attempt**

After f05cf05, all four fresh-reinit codes route through the `else` branch here, which still copies `tail_lines` and `since_time` from the original `stream_logs(...)` call onto each fresh init.

A caller invoking `stream_logs(follow=True, tail_lines=100)` will see the last-100-lines window re-emitted on every SESSION_NOT_FOUND / REPLAY_GAP / RUNNER_* fallback. With `since_time` set, the second init replays everything since the original timestamp. The inline comment at the in-band error branch describes the intent as "the live tail resumes from the current head with no replay of older bytes" — which the request_generator does not honor.

**Suggestion:** Track first-attempt vs subsequent-attempt explicitly (e.g. an `is_first_attempt = attempt == 0` flag, or pass it into the request generator alongside `resume_id` / `resume_off`). Send `tail_lines`/`since_time` only on the very first attempt; subsequent fresh-init retries should omit them and let the server start from the live head. `timestamps` is formatting and can stay across attempts. Worth a unit test in the new `TestStreamLogsResumeStateMachine`: emit a fresh-fallback in `follow=True, tail_lines=N` mode and assert the second init does NOT carry `tail_lines`.

_Found by: reviewer-correctness, reviewer-testing_

src/cwsandbox/_sandbox.py
```diff
@@ -2914,130 +2993,309 @@ async def _stream_logs_async(
...
+
+                    try:
+                        while True:
+                            item = await response_queue.get()
+                            if item is None:
+                                # Stream ended without a terminal frame and
+                                # without a transport error — treat as done.
+                                break
+                            if isinstance(item, grpc.aio.AioRpcError):
+                                if follow and session_id and _is_resumable_transport_error(item):
```
NavarrePratt: [via Claude] **Medium: First transport error is never retried because `session_id` is not set until the first frame arrives**

The guard requires `follow and session_id and _is_resumable_transport_error(item)`. `session_id` is only set when a `LogStreamData` frame arrives, so a transport error during initial connect — or before any frame is received — skips the retry path and surfaces fatally.

The PR description says "on a transient transport error, reconnects with those fields so the live tail continues without the caller seeing the disconnect." That promise holds *after* the first frame but not at the connection edge — a flaky gateway connection at the moment the user opens the tail gives them an immediate error.

The `attempt+1<MAX_ATTEMPTS` gate was correctly dropped in f05cf05 (good); only the `session_id` guard is now over-restrictive.

**Suggestion:** Allow a bounded fresh-init retry on transport error even when `session_id == ""`. The outer `while attempt < MAX_ATTEMPTS` cap already prevents tight-loop pathology. The fresh-init branch is already structured to handle the empty-session case (`if resume_id:` in the request_generator is False → falls into the fresh-init else branch).

_Found by: reviewer-correctness_

src/cwsandbox/_sandbox.py
```diff
@@ -2914,130 +2993,309 @@ async def _stream_logs_async(
...
+                    await output_queue.put(synthesized)
+                    return
 
-                # Flush any remaining partial line
+                # Flush any remaining partial line on clean completion.
                 if line_parts:
                     await output_queue.put("".join(line_parts))
-            except asyncio.CancelledError:
-                raise
             finally:
```
NavarrePratt, line 3356: [via Claude] **Medium: Exceptions raised inside the retry block are silently swallowed behind the EOF sentinel**

The outer `finally` at this line unconditionally enqueues `None` (the EOF sentinel). The outer `except Exception` that follows then enqueues the exception — but `StreamReader` stops iteration on the first `None`, so any exception that arrives *after* the sentinel is silently dropped. The consumer sees a clean end-of-stream with no error surfaced.

The in-code comment at the `except` block claims *"StreamReader stops iteration on Exception, so no trailing None sentinel is needed"* — that is true, but the surrounding control flow does not honor it: the `finally` always emits `None` *first*, then the `except` emits the exception *second*.

Triggering scenarios:
- An asyncio task cancellation propagating mid-loop
- An exception from `stub.StreamLogs(...)` on the second attempt (e.g., a `RuntimeError` from a torn-down channel — related to the channel-reuse comment)
- Any per-attempt setup failure that is not an `AioRpcError`

**Suggestion:** Either enqueue the exception *before* the `None` sentinel, or only enqueue `None` on the success path (move the `None` put into the success branch; let the `except` clause emit only the exception).

_Found by: reviewer-correctness_

src/cwsandbox/_sandbox.py
```diff
@@ -2914,130 +2993,309 @@ async def _stream_logs_async(
...
+                        return request_generator
+
+                    # For follow=False, apply a client-side deadline so a
+                    # stalled server doesn't block the client indefinitely.
+                    # follow=True streams are intentionally unbounded.
+                    grpc_timeout = timeout_seconds if not follow else None
+
+                    call: grpc.aio.StreamStreamCall[
+                        streaming_pb2.LogStreamRequest, streaming_pb2.LogStreamResponse
+                    ] = stub.StreamLogs(
```
NavarrePratt, line 3130: [via Claude] **Medium: `stub` acquired once outside the retry loop and reused on every retry — fragile if the underlying channel is torn down**

`channel` and `stub` are built once before the retry loop (line ~2990, just before `_stream_logs_async`'s main `try`) and reused on every attempt via this `stub.StreamLogs(...)` call. gRPC aio channels usually recover from socket failures transparently via subchannel reconnect, so this works today — but it sets up a latent fragility: if anything tears the channel down between attempts (the integration test `test_stream_logs_reconnects_after_transient_disconnect` does exactly that to force a disconnect), the retry path silently calls `StreamLogs` on a dead stub.

The e2e test passes today only because the channel happens to be recreated by something *outside* the retry loop (likely fresh-init fallback going through a path that touches `_get_or_create_streaming_channel`). A future refactor that resets the streaming channel during error recovery — or any deliberate channel teardown between attempts — will silently break here.

**Suggestion:** Reacquire the stub via `_prepare_streaming_call()` at the top of each attempt (or document the channel-lifetime invariant inline and lock it with a unit test that asserts the second `StreamLogs` runs against a live channel). Cheap to do; eliminates the coupling between retry correctness and gRPC subchannel behavior.

_Found by: reviewer-correctness, reviewer-testing_

src/cwsandbox/_sandbox.py
```diff
@@ -355,6 +358,73 @@ def _translate_rpc_error(
...
+_STREAMING_RESUMABLE_STATUS_CODES: frozenset[grpc.StatusCode] = frozenset(
+    {
+        grpc.StatusCode.UNAVAILABLE,
+        grpc.StatusCode.INTERNAL,
+        grpc.StatusCode.UNKNOWN,
+    }
+)
+
+
+def _is_resumable_transport_error(exc: BaseException) -> bool:
```
NavarrePratt, line 415: [via Claude] **Low (forward-compat): Streaming classifier bypasses AIP-193 reason metadata**

The CANCELLED case from the original review is resolved (good — f05cf05 removed it from `_STREAMING_RESUMABLE_STATUS_CODES` with a clear rationale in the commit message). One structural concern remains:

`_is_resumable_transport_error` dispatches on raw `grpc.StatusCode` only. It does not consult `_translate_rpc_error` (the canonical translator earlier in this file), which folds raw code + AIP-193 ErrorInfo reason into typed `CWSandboxError` subclasses. This means any backend AIP-193 reason attached to a streaming gRPC error (e.g., a hypothetical `CWSANDBOX_RUNNER_TERMINATED` on `INTERNAL`) is invisible to the streaming retry path — the classifier would say "resumable" while `_translate_rpc_error` would map it to a fatal typed exception.

Not a live bug today (the backend does not appear to attach `CWSANDBOX_*` reasons to streaming errors), but it is the kind of forward-compat hazard worth a small inline note so the next reviewer notices when the backend evolves.

**Suggestion:** Either run `_translate_rpc_error` first and dispatch on the resulting typed exception (matches the pattern in `_classify_poll_error` documented in `src/cwsandbox/CLAUDE.md`), or add a comment here noting that this classifier intentionally diverges from the typed-exception classifier for streaming-specific reasons.

_Found by: reviewer-architecture_

NavarrePratt (approved):

PR #133, opened 2026-05-29
feat: typed CWSANDBOX_FILE_TOO_LARGE handling + streaming file APIs

brandonrjacobs:
## Summary

Adds typed handling and streaming APIs for large file transfers. Previously, file operations on payloads larger than the per-file size limit failed with a generic error and no working path for mid-size files. This PR adds a transparent streaming fallback for mid-size files, a typed error for oversized payloads, and dedicated streaming APIs for very large files. Small-file behavior is unchanged.

## Behavior

`read_file` / `write_file` by payload size:

| Size | Behavior |
|---|---|
| small | direct transfer (unchanged) |
| mid-size | transparently streamed; the first fallback per `Sandbox` logs once at `INFO` naming the explicit streaming method |
| very large | refused with a typed `SandboxFileError(CWSANDBOX_FILE_TOO_LARGE)` pointing at the streaming APIs |

## New public surface

- **`CWSANDBOX_FILE_TOO_LARGE`** error reason, surfaced as `SandboxFileError` with `.filepath`, `.reason`, and structured `.metadata` (`size_bytes`, `max_size_bytes`, `operation`). `operation` is the public method name, so callers can branch on it.
- **`read_file_streaming(filepath) -> StreamReader[bytes]`** — chunked reads without buffering the whole payload. After the stream finishes, the SDK verifies the delivered size against the file's size and raises `SandboxFileError(CWSANDBOX_FILE_TOO_LARGE)` on a short read, so a partial read can't be consumed as a complete file.
- **`write_file_streaming(filepath, source)`** — accepts `bytes`, a sync `Iterable[bytes]`, or an `AsyncIterable[bytes]`, streamed without materializing the payload. Non-bytes-like chunks raise `TypeError`.
- **`SandboxStreamBackpressureError`** (a `SandboxExecutionError`) — raised when output is produced faster than the consumer reads it and the stream is ended early. `read_file`, `write_file`, and the explicit `*_streaming` methods all surface this typed error; it is never remasked as a generic error. See `examples/large_file_streaming.py` for the fast-drain pattern.

## Notes

- In the mid-size band the only signal that fallback fired is a one-shot `INFO` log per `Sandbox`. For deterministic streaming, call the explicit `*_streaming` methods.
- Read docstrings document the size bands and the exceptions each method can raise.

## Validation

`ruff` + `mypy` clean; full unit suite green; verified against staging across the size bands (round-trip SHA-256 match for streaming writes).


src/cwsandbox/_sandbox.py
```diff
@@ -4298,8 +4372,88 @@ async def _read_file_via_exec_streaming(self, filepath: str, timeout: float) ->
...
+        - ``delivered >= expected`` (incl. a concurrent append growing the file
+          between the read and the stat): not a truncation, so no raise.
+        """
+        # `stat` is an O(1) metadata lookup, so cap it well below the read's own
+        # (possibly multi-minute) timeout: a slow/wedged channel must not let
+        # the post-read integrity check double the request's wall-clock budget.
+        # Never exceed the caller's remaining timeout.
+        stat_timeout = min(STAT_INTEGRITY_TIMEOUT_SECONDS, timeout)
+        expected = await self._stat_file_size_async(filepath, stat_timeout)
+        if expected is None or expected == 0 or delivered >= expected:
```
djenriquez, line 4539: `_verify_no_truncation` stats the file after the read completes, so a file that grows between cat's EOF and the stat makes `expected` larger than what was legitimately delivered, and `delivered < expected` raises a spurious `CWSANDBOX_FILE_TOO_LARGE`. This fires on the default `read_file()` 32-256 MiB auto-fallback, not just `read_file_streaming`, so `read_file('/var/log/app.log')` on an actively-appended log hard-fails despite receiving a valid complete prefix. In prod the lossless gate is off, so this check is the only truncation defense and the false positive sits on the critical path. Stat before the read and treat post-read growth as benign (`delivered >= stat_before` means no raise) keeps #1172 detection without the false positive. The `delivered >= expected` docstring justification ("concurrent append ... no raise") is also inverted: an append makes `expected` exceed `delivered`, which is the raise branch.

src/cwsandbox/_error_info.py
```diff
@@ -44,13 +44,15 @@
 CWSANDBOX_FILE_IS_DIRECTORY = "CWSANDBOX_FILE_IS_DIRECTORY"
 CWSANDBOX_FILE_IO_FAILED = "CWSANDBOX_FILE_IO_FAILED"
 CWSANDBOX_FILE_PERMISSION_DENIED = "CWSANDBOX_FILE_PERMISSION_DENIED"
+CWSANDBOX_FILE_TOO_LARGE = "CWSANDBOX_FILE_TOO_LARGE"
```
djenriquez, line 51: This reason now covers two opposite conditions: a size-policy refusal where no data was lost and the fix is to switch to streaming, and a post-hoc short read where data was lost and streaming is already in use. Callers can only separate them by sniffing for a `bytes_delivered` metadata key, and the truncation message even tells `read_file_streaming` callers to "use read_file_streaming," a no-op for them. The reason is new in this PR, so splitting the truncation case into its own reason (e.g. `CWSANDBOX_FILE_TRUNCATED`) is free now and a breaking change to the public reason contract after release. Worth splitting before it ships?

src/cwsandbox/_sandbox.py
```diff
@@ -4351,6 +4519,22 @@ def read_file(
         Returns:
             OperationRef[bytes]: Use .result() to block and retrieve contents.
 
+        Behavior:
+            Files up to ~32 MiB are read in a single unary call. Larger files
+            (up to ~256 MiB) transparently fall back to a streaming read — the
+            first such fallback per Sandbox logs once at INFO. Files above
```
djenriquez: This promises files above ~256 MiB are refused, but the `SandboxResourceExhaustedError` fallback below streams with no size check and returns the whole payload in memory, so a multi-GB file rejected as RESOURCE_EXHAUSTED instead of FILE_TOO_LARGE silently buffers until the process OOMs. The uncapped-buffer behavior is pre-existing (that branch's logic is unchanged here), so the in-scope fix is to soften this line to match what the code can actually enforce, since the client can't know remote size on that path. The underlying OOM is worth a separate issue.

src/cwsandbox/_sandbox.py
```diff
@@ -4524,6 +4829,301 @@ def write_file(
...
+            if isinstance(source, (bytes, bytearray, memoryview)):
+                buf = source if isinstance(source, bytes) else bytes(source)
+                for i in range(0, len(buf), STREAMING_WRITE_CHUNK_SIZE):
+                    yield buf[i : i + STREAMING_WRITE_CHUNK_SIZE]
+                return
+            if isinstance(source, AsyncIterable):
+                async for chunk in source:
+                    yield _coerce_bytes_chunk(chunk)
+                return
+            for chunk in source:
```
djenriquez: This consumes a synchronous iterable by calling `next()` directly on the shared background event loop, so a blocking source (a file handle, an NFS/FUSE read, a network generator) stalls every other operation on that loop, including heartbeats and other sandboxes' RPCs. That's the large-upload case this API targets, and the docstring's `open(...)` suggestion leads straight into it. Pumping the sync iterator through `run_in_executor` into a bounded queue keeps it off the loop; `bytes` and `AsyncIterable` sources are already fine.

src/cwsandbox/_sandbox.py
```diff
@@ -4298,8 +4372,88 @@ async def _read_file_via_exec_streaming(self, filepath: str, timeout: float) ->
...
+          report size 0 while ``cat`` legitimately yields content): skip, to
+          avoid a false-positive on a fully-delivered read.
+        - ``delivered >= expected`` (incl. a concurrent append growing the file
+          between the read and the stat): not a truncation, so no raise.
+        """
+        # `stat` is an O(1) metadata lookup, so cap it well below the read's own
+        # (possibly multi-minute) timeout: a slow/wedged channel must not let
+        # the post-read integrity check double the request's wall-clock budget.
+        # Never exceed the caller's remaining timeout.
+        stat_timeout = min(STAT_INTEGRITY_TIMEOUT_SECONDS, timeout)
```
djenriquez: `timeout` here is the original caller budget, not the time remaining after the transfer, so a 600s read that spends 599s streaming can still spend up to 10s more on the stat and blow past the caller's deadline. The constant's comment and this method's docstring both claim it's "capped at the caller's remaining timeout so it never exceeds it," which the code doesn't do. Track an absolute deadline at op start and pass the remaining budget in.

src/cwsandbox/_sandbox.py
```diff
@@ -394,6 +427,36 @@ def _translate_rpc_error(
...
+            "Output stream ended early because it was not being read fast "
+            "enough to keep up with the command's output; some output was "
+            "lost. If you do slow work between reads, move it off the read "
+            "loop (drain into a fast local sink such as a file, then process "
+            "afterward) and use read_file_streaming / write_file_streaming for "
+            "large files. If the destination is itself slow (rate-limited API, "
+            "slow disk) and cannot keep up no matter how tight the loop, split "
+            "the work into smaller transfers. Retrying the same pattern will "
+            "hit this again.",
+            reason=code,
```
djenriquez: `STREAM_BACKPRESSURE` rides on the same `reason=` field that otherwise carries AIP-193 `CWSANDBOX_*` reasons, but it's an ExecStreamError stream code, not an ErrorInfo reason (the comment in `_error_info.py` says as much). Callers branching on `.reason` can no longer assume it's an AIP reason. A separate attribute like `stream_code`, or documenting `.reason` as "any programmatic code," keeps the namespaces from blurring. Intentional, or worth separating?

src/cwsandbox/_sandbox.py
```diff
@@ -4444,6 +4641,74 @@ async def _write_file_via_exec_streaming(
...
+        meta = exc.metadata or {}
+        raw = meta.get("max_size_bytes")
+        if not raw:
+            return
+        try:
+            value = int(raw)
+        except (TypeError, ValueError):
+            return
+        if value > 0:
+            self._observed_file_op_cap_bytes = value
```
djenriquez, line 4788: The observed cap is cached with only a `value > 0` guard and no upper clamp, and the old `DEFAULT_FILE_UNARY_SAFE_LIMIT_BYTES` that kept unary payloads under the gRPC frame limit was deleted. A cluster reporting a cap near or above the 100 MiB frame ceiling would route a sub-cap payload to the unary path and hit a frame-size reject. This cache also only ever lowers the cap, so on a cluster with an operator-raised cap (the server allows up to ~96 MiB) the hardcoded 32 MiB still sends valid unary ops to streaming, since no rejection occurs to populate it. Clamping to a frame-safe ceiling, or seeding from server-advertised config, covers both.

src/cwsandbox/_sandbox.py
```diff
@@ -4298,8 +4372,88 @@ async def _read_file_via_exec_streaming(self, filepath: str, timeout: float) ->
...
+        if returncode != 0:
+            return None
+        text = stdout.decode("utf-8", errors="replace").strip()
+        try:
+            value = int(text)
+        except ValueError:
+            return None
+        return value if value >= 0 else None
+
+    async def _verify_no_truncation(
```
djenriquez: Every streamed or fallback read now issues a second StreamExec to run `stat`, a fixed extra round-trip plus a sandbox process spawn on each read, including the 32-99 MiB band where the >262 MiB truncation bug can't occur. It's O(1) metadata so it's negligible against a large transfer, but gating the check by size band or backend version would avoid paying it where it can't help. Worth gating?

djenriquez (commented):
Looked closely at the large-file dispatch and the new streaming paths. A couple of items I'd want to settle before merge; the rest are smaller flags.

src/cwsandbox/_sandbox.py
brandonrjacobs: Fixed in 9396d21. The file's size is now captured **before** the read instead of after, so a concurrent append can't make `expected` exceed what was legitimately delivered. For the `read_file` auto-fallback specifically there's no stat at all — it reuses the server-reported `size_bytes` from the `FILE_TOO_LARGE` metadata that triggered the fallback as the pre-read baseline. `_verify_no_truncation` is now a pure comparison taking `expected` as a parameter; `delivered >= expected` (including a growing file) never raises. The inverted docstring justification is corrected: an append makes `delivered` exceed the *pre-read* baseline, which is the no-raise branch. Added `test_read_fallback_growing_file_no_false_positive` to pin the appended-log case.

src/cwsandbox/_error_info.py
brandonrjacobs: Agreed — split before it ships. 9396d21 adds a distinct `CWSANDBOX_FILE_TRUNCATED` reason for the post-hoc short-read case (data lost, already streaming), leaving `CWSANDBOX_FILE_TOO_LARGE` for the size-policy refusal (no data lost, switch to streaming). Both are in `FILE_ERROR_REASONS`. The truncation message keeps the "use read_file_streaming / read in smaller parts" guidance, which is still the right advice for the `read_file` exec-fallback path that emits it (that path is *not* `read_file_streaming` — it's the unary-cap fallback). Callers can now branch on `reason` without sniffing `bytes_delivered`.

src/cwsandbox/_sandbox.py
brandonrjacobs: Softened in 9396d21. The docstring no longer claims all >256 MiB reads are refused — it now scopes the refusal to "when the server reports the file's size" and explicitly calls out that on the `RESOURCE_EXHAUSTED` path the client can't know the remote size, so a very large file may be buffered in full rather than refused, pointing callers at `read_file_streaming`. I've left the underlying uncapped-buffer OOM on that branch as-is (pre-existing, that branch's logic is unchanged here) and will file a separate issue for it.

src/cwsandbox/_sandbox.py
brandonrjacobs: Fixed in 9396d21. The sync-iterable branch now pulls each chunk via `loop.run_in_executor(...)` through a small async generator (`_iter_sync_source_in_executor`), so a blocking `next()` parks an executor thread instead of the shared loop. The iterator advances exactly one step per consumed item, so no prefetch buffer / hand-off race is needed and the downstream exec stdin queue still applies backpressure. `bytes` and `AsyncIterable` sources are untouched. The `TypeError` contract for bad chunks is preserved (raised in the worker thread, carried back on the Future, re-raised unchanged). Added `test_write_streaming_sync_iterable_runs_off_event_loop`, which blocks the source on a `threading.Event` and asserts a concurrently-scheduled coroutine still makes progress. Updated the docstring caveat accordingly.

src/cwsandbox/_sandbox.py
brandonrjacobs: Fixed in 9396d21. The streaming read now records an absolute `time.monotonic()` deadline at op start. The pre-read stat draws from the **remaining** budget (`_stat_budget` → `min(STAT_INTEGRITY_TIMEOUT_SECONDS, remaining)`), and the read that follows gets `_remaining_budget(deadline)` after the stat — so stat + read together honor the caller's overall timeout. Because the stat now runs *before* the read (the false-positive fix), it can't be tacked on after a long read at all. The constant comment and docstring are reworded to match. Added `test_stat_uses_remaining_budget_not_full_timeout` (asserts a 600s caller timeout still caps the stat at the ceiling) and `test_remaining_budget_helpers`.

src/cwsandbox/_sandbox.py
brandonrjacobs: Good call — separated. 9396d21 adds a dedicated `stream_code` attribute to `SandboxStreamBackpressureError` and stops overloading `.reason`: `_exec_stream_error` now passes `stream_code=code`, and the exception's `__init__` does not accept/forward `reason` (it stays `None`). So `.reason` remains purely the AIP-193 ErrorInfo namespace; callers branch on the exception class (or `stream_code`). The class still subclasses `SandboxExecutionError`, so existing handlers are unaffected. Added `TestSandboxStreamBackpressureError` covering the namespace split and that `reason=` is rejected.

src/cwsandbox/_sandbox.py
brandonrjacobs: Fixed in 9396d21. Restored the frame-safe ceiling as `MAX_FILE_UNARY_BYTES` (channel max message length − 1 MiB framing headroom) and clamp the effective cap to it in `_file_op_cap`, so even a cluster reporting a cap at/above the frame limit can't route a sub-cap payload to the unary path and hit a frame reject — anything above the clamp streams. The raw observed value is still stored faithfully; the clamp is applied at the point of use. `test_observed_cap_clamped_to_frame_safe_ceiling` pins this.

On the "cache only ever lowers" half (operator-*raised* cap → valid larger unary ops needlessly stream): that direction is conservative-safe — streaming a payload that *could* have gone unary is correct, just a hair less optimal — and the SDK has no server-advertised config to seed from today. I've left the 32 MiB default as-is rather than guess upward; happy to revisit if/when the backend exposes the cap in a discovery response.

src/cwsandbox/_sandbox.py
brandonrjacobs: Gated in 9396d21, two ways:

1. The `read_file` exec-fallback path (the 32–99 MiB band you cite) no longer issues a stat at all — it reuses the server-reported `size_bytes` from the `FILE_TOO_LARGE` metadata as the pre-read baseline. Zero extra round-trips there.
2. `read_file_streaming` (no prior size signal) gates the check on `TRUNCATION_CHECK_MIN_BYTES` (½ the auto-fallback ceiling): silent truncation is a large-payload phenomenon, so for a small file the comparison is skipped. The one O(1) stat on the genuinely-large-file API is negligible against the transfer it guards.

`test_small_file_skips_stat_round_trip` confirms a small streamed read returns cleanly even when delivered < stat size (band gate skips it).

src/cwsandbox/_sandbox.py
brandonrjacobs: Filed the OOM follow-up as #135.

brandonrjacobs:
Thanks for the close review @djenriquez — addressed all eight in 9396d21 (replies on each thread). Summary:

**Blockers**
- **Truncation false-positive** → stat the size *before* the read; the `read_file` fallback reuses the server-reported size (no extra stat), `read_file_streaming` stats once up front. A concurrent append can no longer look like a short read.
- **stat timeout overshoot** → absolute monotonic deadline at op start; the pre-read stat draws from the *remaining* budget capped at the integrity ceiling, the read gets what's left.
- **Sync iterable on the loop** → each `next()` now runs in the default executor, so a blocking source parks a worker thread, not the shared loop.
- **Unclamped observed cap** → restored a frame-safe ceiling (`MAX_FILE_UNARY_BYTES`) and clamp to it, so a too-high reported cap can't route a sub-cap payload into a frame-size reject.

**Reason split**
- New `CWSANDBOX_FILE_TRUNCATED` for the post-hoc short-read case, separate from the `CWSANDBOX_FILE_TOO_LARGE` size-policy refusal — done now while unreleased.

**Your two questions**
- `STREAM_BACKPRESSURE` moved off `.reason` onto a dedicated `.stream_code` attribute (kept out of the AIP-193 namespace).
- Stat integrity check now gated by size band; the `read_file` fallback issues no stat at all.

**Docs / follow-up**
- `read_file` docstring no longer over-promises refusal on the `RESOURCE_EXHAUSTED` path; the underlying uncapped-buffer OOM is tracked in #135.

`ruff` + `mypy` clean; full unit suite green (1171 tests, +9 new).

djenriquez (approved):
LGTM

PR #134, opened 2026-05-29
fix: fall back to exec-streaming on FAILED_PRECONDITION

nicholaspun-wandb:
## Summary

- Add `SandboxFailedPreconditionError` and map gRPC `FAILED_PRECONDITION` to it in `_translate_rpc_error`
- Catch the new exception inside `_write_file_async` and route to the existing exec-streaming fallback so oversize writes succeed instead of failing with an opaque `SandboxError`

## Why

The gateway emits `FAILED_PRECONDITION` with AIP-193 reason `CWSANDBOX_FILE_TOO_LARGE` whenever an `AddFile` payload exceeds the configured `max-file-operation-bytes` cap (32 MiB by default — see [aviato config.go:310](https://github.com/coreweave/aviato/blob/main/internal/config/config.go#L310)). The SDK's existing fallback only catches `RESOURCE_EXHAUSTED` with `"larger than max"` wording, which is gRPC's *client-side envelope* error, not the gateway's *server-side payload-cap* error.

The two limits are independent:

- gRPC envelope: `DEFAULT_GRPC_MAX_MESSAGE_LENGTH_BYTES = 100 MiB`; SDK pre-dispatches to exec-streaming at 99 MiB.
- Server payload cap: `DefaultMaxFileOperationBytes = 32 MiB` (configurable per deploy; can go up to ~96 MiB).

Any file size between the server cap (typically 32 MiB) and the SDK's pre-dispatch threshold (99 MiB) currently falls through to a generic `SandboxError("Write file failed: file payload exceeds configured max-file-operation-bytes")`. The structured `max_size_bytes` metadata the gateway attaches is dropped on the floor, and the user has no easy way to know they hit a cap they can simply chunk around.

This change routes those writes through the same exec-streaming path that already handles the gRPC envelope case (chunks via stdin, bypasses the unary cap). After the fix, write_file at 33–96 MiB just works.

## Test plan

- [x] `mise run typecheck` — clean
- [x] `mise run lint` — clean
- [x] `mise run test` — 1136 unit tests pass (+2 new)
- [ ] Minimal repro succeeeds:
```python
import os
from wandb.sandbox import Sandbox

SIZE = 33 * 1024 * 1024
payload = os.urandom(SIZE)
print(f"Attempting write_file of {SIZE / (1024 * 1024):.1f} MiB ...")
with Sandbox.run(container_image="python:3.11", max_lifetime_seconds=120) as sb:
    sb.wait(timeout=120)
    print(f"Sandbox: {sb.sandbox_id}")
    sb.write_file("/tmp/big.bin", payload, timeout_seconds=60).result()
```

Before:
```bash
uv run python repro_file_cap.py
Attempting write_file of 33.0 MiB ...
wandb: [sandbox] Loaded credentials for https://api.wandb.ai from /Users/nicholaspun/.netrc.
Sandbox: d6d98a7c-91ec-44c0-8442-42060d9a566e
Traceback (most recent call last):
  File "/Users/nicholaspun/Documents/test-wb-sandbox/.venv/lib/python3.11/site-packages/cwsandbox/_sandbox.py", line 4378, in _write_file_unary_async
    response = await self._stub.AddFile(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nicholaspun/Documents/test-wb-sandbox/.venv/lib/python3.11/site-packages/grpc/aio/_call.py", line 329, in __await__
    raise _create_rpc_error(
grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
	status = StatusCode.FAILED_PRECONDITION
	details = "file payload exceeds configured max-file-operation-bytes"
	debug_error_string = "UNKNOWN:Error received from peer ipv4:166.19.70.19:443 {grpc_status:9, grpc_message:"file payload exceeds configured max-file-operation-bytes"}"
>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/nicholaspun/Documents/test-wb-sandbox/repro_file_cap.py", line 13, in <module>
    sb.write_file("/tmp/big.bin", payload, timeout_seconds=60).result()
  File "/Users/nicholaspun/Documents/test-wb-sandbox/.venv/lib/python3.11/site-packages/cwsandbox/_types.py", line 78, in result
    return self._future.result(timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nicholaspun/.local/share/uv/python/cpython-3.11.11-macos-aarch64-none/lib/python3.11/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/nicholaspun/.local/share/uv/python/cpython-3.11.11-macos-aarch64-none/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/Users/nicholaspun/Documents/test-wb-sandbox/.venv/lib/python3.11/site-packages/cwsandbox/_sandbox.py", line 4484, in _write_file_async
    await self._write_file_unary_async(filepath, contents, timeout)
  File "/Users/nicholaspun/Documents/test-wb-sandbox/.venv/lib/python3.11/site-packages/cwsandbox/_sandbox.py", line 4382, in _write_file_unary_async
    raise _translate_rpc_error(
cwsandbox.exceptions.SandboxError: Write file failed: file payload exceeds configured max-file-operation-bytes
```

After:
```bash
```

## Discovered while

Stress-testing write_file throughput against `wandb-prod-east-14`. Hit the cap at 33 MiB on `write_file()` despite the SDK pre-dispatch threshold being 99 MiB. Tracking the wrong limit was the root cause.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nicholaspun-wandb:
resolved with https://github.com/coreweave/cwsandbox-client/pull/133

PR #138, opened 2026-06-09
fix: skip signal handlers outside main thread (#136)

### LINKED ISSUE #136: Import time signal handler install crashes from non-main threads such as marimo
### Description

`cwsandbox/_cleanup.py` calls `_install_handlers()` at module load, which calls `signal.signal()`. That only works on the main thread, so simply importing `cwsandbox` from a worker thread crashes with `ValueError: signal only works in main thread of the main interpreter`.

```python
import threading
threading.Thread(target=lambda: __import__("cwsandbox")).start()
# ValueError: signal only works in main thread of the main interpreter
  ```
marimo executes the with `app.setup:` block from a worker thread, so any marimo notebook that imports cwsandbox in setup fails before the first cell renders.

This is not a hard blocker. For now I have a workaround in my marimo notebook that catches the ValueError and restores the original just wanted to file this in case it impacts others 

### Steps to reproduce

```
import marimo
app = marimo.App()

with app.setup:
    from cwsandbox import Sandbox  
```

### Expected behavior

Import should succeed. 

### Actual behavior

`ValueError: signal only works in main thread of the main interpreter`

### cwsandbox version

0.24.0

### Python version

3.13.9

### Operating system

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by raywcm**
Closes #136

## Summary

Importing `cwsandbox` from a worker thread currently fails because `_cleanup.py` installs SIGINT/SIGTERM handlers at import time, and `signal.signal()` is only valid in the main thread.

This keeps the `atexit` cleanup registration available from any thread, but skips installing signal handlers when `_install_handlers()` runs outside the main thread. If the module is later initialized from the main thread, the signal handlers can still be installed.

## Verification

- Added unit coverage for worker-thread handler installation.
- Confirmed a worker-thread import of `cwsandbox._cleanup` no longer raises `ValueError`.
- Ran `pytest tests/unit/cwsandbox/test_cleanup.py -q`: 18 passed.
- Ran `python -m ruff check src/cwsandbox/_cleanup.py tests/unit/cwsandbox/test_cleanup.py`: all checks passed.

### DIFF
--- a/src/cwsandbox/_cleanup.py
+++ b/src/cwsandbox/_cleanup.py
@@ -17,6 +17,7 @@ import atexit
 import logging
 import signal
 import sys
+import threading
 from collections.abc import Callable
 from types import FrameType
 
@@ -30,6 +31,7 @@ _cleanup_in_progress: bool = False
 _original_sigint: _SignalHandler = None
 _original_sigterm: _SignalHandler = None
 _handlers_installed: bool = False
+_atexit_registered: bool = False
 
 
 def _cleanup() -> None:
@@ -94,13 +96,18 @@ def _install_handlers() -> None:
 
     The original signal handlers are preserved and chained after cleanup.
     """
-    global _original_sigint, _original_sigterm, _handlers_installed
+    global _original_sigint, _original_sigterm, _handlers_installed, _atexit_registered
+
+    if not _atexit_registered:
+        atexit.register(_cleanup)
+        _atexit_registered = True
 
     if _handlers_installed:
         return
 
-    # Register atexit handler
-    atexit.register(_cleanup)
+    if threading.current_thread() is not threading.main_thread():
+        logger.debug("Skipping signal handlers outside the main thread")
+        return
 
     # Install signal handlers, preserving originals for chaining
     _original_sigint = signal.signal(signal.SIGINT, _signal_handler)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by brandonrjacobs — COMMENTED** [review]
Thanks for fixing the crash — the main-thread guard is the right call given signal.signal()'s constraint. One gap worth mentioning though:
  
  The signal handlers are never recovered after a worker-thread-first import. 
```
_install_handlers() runs only once, at import (_cleanup.py:135), and nothing else
  in src/ ever calls it again. So in the exact scenario this PR targets — first import cwsandbox happening on a non-main thread — the signal handlers are
  skipped and stay skipped for the life of the process. A re-import from the main thread is a cached no-op and doesn't re-run the install.
```

  Concretely: a later SIGTERM (e.g. a k8s/orchestrator shutdown) terminates the process without running _cleanup — atexit doesn't fire on signal-induced
  termination — so sandboxes get orphaned and keep consuming compute. That's the guarantee this module exists to provide.

  I verified this with a runtime repro of the control flow: after a worker-thread install, _handlers_installed=False and SIGTERM/SIGINT are still at default
  disposition; only an explicit main-thread call to _install_handlers() ever wires them up.

  Two smaller notes:
  - The PR description says handlers "can still be installed" from the main thread later — that recovery path doesn't actually exist in the code.
  - The skip is logged at debug, so this degradation is silent. Worth bumping to warning.
  
  Suggestion: the root cause is doing main-thread-only work (signal.signal) as an import-time side effect. Rather than installing at import, make installation lazy + idempotent and trigger it from the first real SDK use (e.g. _LoopManager.get() / Sandbox.run()). That attempts install on the main thread whenever the SDK is actually used, which delivers the recovery this PR describes — and ends up being less code than the current flag-juggling. Happy to discuss.

PR #146, opened 2026-08-12
feat!: cut over public SDK to Sandbox v1 API (1.0.0)

djenriquez:
## Summary

This release speaks only the Sandbox v1 control plane. Callers that still need v1beta2 should stay on 0.26.x. The sync/async hybrid client shape is unchanged: create, exec, files, logs, sessions, and remote functions still work the same way from the caller's seat, against the new wire protocol and placement/network model. Create can optionally retry once on the other placement pool when the first attempt fails for capacity or placement constraints. Empty beta echo fields for profile and network modes are not part of the public surface; use per-service URLs instead.

## What changed

**Wire protocol and stubs**
Vendors consumer stubs for sandbox, discovery, settings, templates, and volumes under `src/cwsandbox/_proto/`. Removes beta gateway/streaming/secrets stubs. `scripts/update-protos.sh` and `scripts/buf.gen.python.yaml` generate against protobuf runtime v26.1.

**Placement and discovery**
Drops profiles from the public API. Runner listing uses capabilities/visibilities; workloads select `placement_mode` and runner filters. Templates cover CKS pod fragments (`run_from_template`).

**Placement spillover**
Adds `placement_spillover` on defaults / `Sandbox.run` / session / `@session.function` (`strict` default, `cks_then_serverless`, `serverless_then_cks`). On a spillable create failure, CreateSandbox retries once with the alternate mode and a new request id; CKS→serverless clears `runner_ids`. Template creates stay `strict` only. Reasons and helpers live in `src/cwsandbox/_error_info.py`; create path in `src/cwsandbox/_sandbox.py`.

**Network and volumes**
Port exposure is typed `services=[Service(...)]`. `NetworkOptions` is deny-flag only. Scratch volumes (`ScratchVolumeOptions`) are the named-mount path for filesystem snapshots; `file_system_snapshot=` remains a convenience. Public URL surface is `service_urls` (and `exposed_ports` derived from services). `profile_id`, `service_address`, `applied_ingress_mode`, and `applied_egress_mode` are removed. CLI `ls` shows runner, not profile.

**Lifecycle mapping**
Create uses a frozen request id for idempotency. `stop()` still exists for callers and maps to DeleteSandbox. Rejects `s3_mount`, `ports`, and `max_timeout_seconds` with `TypeError` (client deadlines stay on `request_timeout_seconds`). List filter is `show_terminated` (not `include_stopped`).

**Version and docs**
Package version is `1.0.0`. AGENTS, examples, and public docstrings match the v1 surface, including a short spillover tip in `examples/discover_infrastructure.py`. Deferred backends that were never public on 0.26 (WIF admin, SecretStore admin, NetworkService, TOKEN endpoints) stay unsupported — no hybrid dialect.

**Tests**
Unit and integration suites target v1 only (profile pin tests removed or rewritten for runners/placement). Spillover coverage in `tests/unit/cwsandbox/test_placement_spillover.py`.

## Behavior / Key decisions

| Input / old surface | Outcome on this branch |
| --- | --- |
| Profile selectors / profile discovery | Removed; use placement + runners (+ templates for CKS) |
| `profile_id` / `service_address` / `applied_*_mode` | Removed; use `service_urls` |
| String network modes / `ports=` | Typed `services=` |
| `s3_mount` / `max_timeout_seconds` | Loud `TypeError` |
| `include_stopped` / `idempotency_key` | `show_terminated` / `request_id` |
| `stop()` | Same ergonomics; DeleteSandbox underneath |
| Capacity / placement reject on create + non-strict spillover | One alternate-mode create retry (new request id; clear runners when spilling to serverless) |
| Template create + non-strict spillover | Rejected |
| Serverless product gates / auth / invalid argument | No spill |
| v1beta2 backends | Unsupported; stay on 0.26.x |
| WIF/SecretStore admin, NetworkService, TOKEN | Unsupported (never public on 0.26 either) |

## Risk surface

- Hard break for every consumer still on v1beta2 or profile-based placement.
- Call sites that read `profile_id`, `service_address`, or `applied_*_mode` must move to `service_urls` / placement.
- Public exports change — docs API-ref generator `MANIFEST_GROUPS` needs a matching update in `coreweave/docs` (include `PlacementSpillover`; drop hollow attrs).
- Spillover is create-time only (not post-accept poll FAILED) and at most one hop.
- CKS template pod attachments are live in prod (aviato 0.147.0 / #1733, from #1731). `run_from_template` is the consumer path; admins author attachments on `SandboxTemplate`.

## Test plan

- [x] `uv run pytest tests/unit -q` — 1146 passed (earlier in cutover)
- [x] `uv run pytest tests/integration -q` against production — 103 passed, 2 skipped (earlier in cutover)
- [x] `uv run pytest tests/unit/cwsandbox/test_placement_spillover.py -q` — 17 passed
- [x] `uv run pytest tests/unit/cwsandbox/test_lifecycle.py tests/unit/cwsandbox/test_cli_list.py tests/unit/cwsandbox/test_sandbox.py tests/unit/cwsandbox/test_sandbox_polling.py tests/unit/cwsandbox/test_placement_spillover.py -q` — 483 passed (hollow-attr removal)
- [x] Docs/examples retargeted to v1 (AGENTS, discover_infrastructure, list_stopped_sandboxes, swebench)
- [ ] Confirm `coreweave/docs` API-ref manifest when marking ready


brandonrjacobs (commented):
Ran this branch end-to-end against production before reviewing: create → get → list-by-tag → exec → file write/read → delete all work cleanly through the new v1 surface (create accepted in ~1s, RUNNING in ~5–7s), all four removed-kwarg `TypeError` guards fire with good messages, `show_terminated` behaves, and typed `Service` exposure with `custom` visibility places and runs fine. The core cutover is in good shape.

Review did turn up one blocking issue and several majors, mostly where the new surface meets the live API. File refs are against 738dec4.

### Blocking

**1. `service_urls` can never be populated — and it's the documented replacement for `service_address`.**
The only place `_service_urls` is filled from a server response is the create echo (`_apply_create_echo`, `src/cwsandbox/_sandbox.py:3902`), but service URLs aren't available at create time — they only exist once the sandbox is RUNNING. Nothing refreshes them afterwards: the poll/Get path (`_apply_sandbox_info`) never reads `status.services`, and `_from_sandbox_info` (`from_id`/`list`) hard-resets `_service_urls = ()` (`src/cwsandbox/_sandbox.py:1919`). Confirmed in production: a RUNNING sandbox with a declared service returns `service_urls == ()` on every `get_status()` poll while `exposed_ports` fills from the create echo. The CHANGELOG and AGENTS.md direct migrating users to this property as the replacement for the removed `service_address`. The one test that could catch it, `test_sandbox_public_service_connectivity` (`tests/integration/cwsandbox/test_sandbox.py:736`), does `pytest.skip` when `service_urls` is empty — so it skips on every run, forever. Fix needs (a) reading `status.services` on the Get/poll paths, and (b) making that test fail rather than skip when URLs never arrive.

### Placement spillover

**2. Spillover doesn't fire on the capacity signal the create path actually emits.**
`SPILLOVER_ELIGIBLE_REASONS` (`src/cwsandbox/_error_info.py:118`) is `{RUNNER_CAPACITY_EXHAUSTED, PLACEMENT_REJECTED, PLACEMENT_CONSTRAINT_UNSATISFIED}`, but pool exhaustion on a v1 create (no eligible runner / runners at capacity) surfaces to clients as `UNAVAILABLE` + `CWSANDBOX_RUNNER_UNAVAILABLE` — which the SDK classifies as non-spillable and also never retries. Net effect: the headline `cks_then_serverless` scenario ("CKS pool full → spill to serverless") raises `SandboxUnavailableError` with no spill, and `placement_spillover` degrades to roughly `strict` outside of constraint-shaped rejects. Worth confirming the exact reason set the create path emits for capacity exhaustion and aligning `SPILLOVER_ELIGIBLE_REASONS` with it — plus a unit test in `test_placement_spillover.py` pinning the intended behavior for the `UNAVAILABLE`/`CWSANDBOX_RUNNER_UNAVAILABLE` shape, whichever way that decision goes.

**3. A failed spillover permanently mutates the Sandbox's placement state.**
`_create_with_optional_spillover` flips `self._placement_mode`, clears `self._runner_ids`, and mints a new request id in place (`src/cwsandbox/_sandbox.py:3564-3579`) with no restore when attempt 2 fails. `start()` is documented idempotent/safe to call multiple times, so a caller that retries after the exception now starts from the flipped state — and because `alternate` is derived from the *policy* rather than the mode being attempted, the retried call "spills" serverless→serverless: the user's configured CKS primary and their `runner_ids` pin are never attempted again. Repro against a mocked stub: two `start()` calls produced create attempts `[CKS, SERVERLESS, SERVERLESS, SERVERLESS]` with three distinct request ids. Suggest snapshotting and restoring the placement fields around the attempt loop (or deriving each attempt's request from immutable originals).

**4. Bare reason-less `RESOURCE_EXHAUSTED` triggers an immediate, silent mode flip.**
`_is_spillover_eligible` (`src/cwsandbox/_sandbox.py:877`) treats any `SandboxResourceExhaustedError` carrying no `CWSANDBOX_*` reason as spill-eligible. That shape can come from throttling layers or client-local gRPC conditions — none of which indicate the target pool is out of capacity — and the flip happens instantly with no backoff, silently changing where the workload lands (and dropping the runner pin on CKS→serverless). `test_bare_resource_exhausted_spills` (`tests/unit/cwsandbox/test_placement_spillover.py:278`) locks this in as contract. Recommend requiring an explicit eligible reason, or at minimum not mode-flipping on reason-less shapes.

### Lifecycle

**5. `stop(missing_ok=True)` now fails for absent sandboxes.**
`_do_stop` sends `allow_missing=missing_ok` on the delete (`src/cwsandbox/_sandbox.py:4563-4568`). Under standard `allow_missing` semantics the delete *succeeds* (OK, empty response) when the sandbox is absent — so the client-side latch at `src/cwsandbox/_sandbox.py:4581` (`if missing_ok and is_not_found(...)`) never fires. The SDK then proceeds into the post-stop poll, `GetSandbox` returns NOT_FOUND, and after the 2s post-stop retry budget it raises `SandboxTerminalStateUnavailableError` — the opposite of the docstring's "ignore if already deleted". The stop-flow unit tests only model a NOT_FOUND-returning server, so they can't see it. Either treat the OK-but-empty delete response as terminal when `missing_ok` was requested, or don't set `allow_missing` and keep relying on the existing NOT_FOUND latch.

**6. `stop()` grace docs and FSS deadline math still describe beta semantics.**
On v1, `grace_period_seconds=0` means immediate termination. The docstring at `src/cwsandbox/_sandbox.py:4785-4791` still says the backend substitutes a ~30s default when 0 is sent, `DEFAULT_FSS_STOP_GRACE_FALLBACK_SECONDS` (`src/cwsandbox/_defaults.py`) budgets the client deadline on that premise, and `tests/unit/cwsandbox/test_file_system_snapshot.py:472` asserts it. Docs, deadline math, and test should move to the v1 contract together.

### Templates / streaming

**7. `run_from_template` sparse container overrides are rejected by the API.**
Container overrides currently replace the template's container list wholesale rather than merging per-field, so the sparse `PartialContainer(name="main")` built by `_build_create_from_template_request` (`src/cwsandbox/_sandbox.py:3849`) fails validation for every partial override — command-only, args-only, env-only, files-only, etc. get `CWSANDBOX_INVALID_REQUEST` ("image is required") or `CWSANDBOX_NOT_IMPLEMENTED` ("args without command"). Either build a complete container override client-side or only attach one when the caller supplied a complete container; and an integration test that overrides just `command` on a real template would catch this class of issue.

**8. Log-stream resume sends `timestamps=true`, which resume rejects — so resume never works with timestamps.**
`_stream_logs_async` (`src/cwsandbox/_sandbox.py:4984-4991`) copies `timestamps` into every attempt's request, including resume attempts that carry `resume_log_session_id`/`resume_log_offset`. Resume requests with `timestamps` set are rejected as `INVALID_ARGUMENT` (formatting is inherited from the original stream), and INVALID_ARGUMENT is not in `_STREAMING_RESUMABLE_STATUS_CODES` — so a `stream_logs(follow=True, timestamps=True)` that hits a transient disconnect dies permanently instead of resuming, exactly the case the resume machinery exists for. The cutover also deleted the resume wire-shape test suite (`TestStreamLogsResumeStateMachine` et al.) that asserted resume request shapes; those tests were beta-specific only in their stub method names and are worth porting rather than dropping.

### API surface / examples

**9. `session.sandbox()` rejects `request_timeout_seconds` — including in the shipped example.**
`examples/swebench/run_evaluation.py:163` (migrated by this PR) passes `request_timeout_seconds` into `session.sandbox(...)`, which raises `TypeError: session.sandbox() got unexpected keyword argument(s)` — the parameter isn't plumbed through `Session.sandbox()`/`session.function()` (`src/cwsandbox/_session.py`). Since the `max_timeout_seconds` TypeError message explicitly tells users to switch to `request_timeout_seconds`, session users currently have no migration path. Suggest plumbing it through Session (Sandbox.run already accepts it), which also fixes the example.

**10. `TestCrossLoopRegression` (8 tests) deleted although the machinery it guarded is unchanged.**
The dual-event-loop routing those tests pinned — `__await__`/`__aenter__` routing through `_loop_manager.run_async` + `asyncio.wrap_future` (`src/cwsandbox/_sandbox.py:4441-4461`) — survives the cutover unchanged, and nothing replaces the coverage (`test_loop_manager.py` covers only the primitives). A future simplification to plain `await` would silently break every multi-loop consumer. The tests only referenced the beta in mocked stub method names; port them (StartSandbox→CreateSandbox) rather than dropping.

### Minor

- `SandboxDefaults.from_dict` docstring promises coercion of `services`/`volumes` lists, but the loop only coerces `args`/`tags`/`runner_ids` (`src/cwsandbox/_defaults.py:437`); `volumes` passed as dicts crash later with `AttributeError` when used via `session.sandbox()`.
- `from_dict` silently drops removed keys (`profile_ids`/`profile_names`) via the unknown-key filter (`src/cwsandbox/_defaults.py:404`) — inconsistent with the loud-`TypeError` contract everywhere else; a 0.26 defaults dict that pinned placement now silently loses that intent.
- `exposed_ports` (`src/cwsandbox/_sandbox.py:3904`) reports declared-but-unexposed (UNSPECIFIED-visibility) ports as exposed.
- `tests/integration/cwsandbox/test_sandbox.py:703` mixes PUBLIC and PRIVATE services in one create, which the API currently rejects — the test can't pass as written.
- Spillover unit fixtures pair reasons with gRPC codes that don't match the error contract (e.g. `CWSANDBOX_PLACEMENT_CONSTRAINT_UNSATISFIED` is FAILED_PRECONDITION-shaped, not RESOURCE_EXHAUSTED; `tests/unit/cwsandbox/test_placement_spillover.py:133`), so the real translate paths go unexercised.
- `examples/discover_infrastructure.py:65` pins `runners[0]` in the CKS snippet without filtering out `is_shared` runners, which can't be pinned in CKS mode.
- Docstring drift: `NetworkOptions` (`src/cwsandbox/_types.py:199`) claims it "declares CUSTOM-visibility ingress sources and egress allow rules" but the class is deny-flags only; `FileSystemSnapshotTrigger` (`src/cwsandbox/_types.py:373`) documents a `STOP` alias this PR removed; `src/cwsandbox/_network.py:11` module docstring still references the deleted beta stubs.

djenriquez (commented):
Landed in 3d398b5. Capacity exhaustion stays `CWSANDBOX_RUNNER_CAPACITY_EXHAUSTED` via aviato#1739; the SDK still only spills on the typed capacity/placement reasons, not `RUNNER_UNAVAILABLE`.

1. `service_urls` — Get/poll/`from_id` refresh from `status.services`. E2e fails if URLs never arrive when PUBLIC runners exist; skips on custom-only fleets.
2. Spillover vs `RUNNER_UNAVAILABLE` — no SDK change; server-side reason fix in aviato#1739.
3. Failed spill mutates placement — restore original mode/`runner_ids`/`request_id` if attempt 2 fails.
4. Bare `RESOURCE_EXHAUSTED` — spill needs an explicit eligible reason.
5. `stop(missing_ok=True)` — empty OK delete counts as terminal; no post-stop poll.
6. Grace / FSS deadline — aligned to v1 (`0` means immediate).
7. Template container overrides — only send a replace when the caller gives a full image (args require command).
8. Log resume + `timestamps` — omit `timestamps` on resume attempts; resume tests restored.
9. Session `request_timeout_seconds` — accepted on `session.sandbox()` and `@session.function()`.
10. Cross-loop regression tests — restored against CreateSandbox.

Same commit also covers the minors: loud `from_dict` profile reject, services/volumes coerce, drop UNSPECIFIED from `exposed_ports`, mixed PUBLIC+PRIVATE e2e expects `CWSANDBOX_NOT_IMPLEMENTED`, spillover fixture codes, discover example filters `is_shared`, docstring cleanup.

Prod e2e on this org: 92 passed, 3 skipped (no PUBLIC/PRIVATE runners), mixed-reject passed.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5033,126 @@ async def _stream_logs_async(
...
-            # draining.
+            if not done:
+                raise SandboxUnavailableError(
+                    f"Log stream for sandbox {sandbox_id} could not be resumed"
+                ) from last_transport_error
+            inner_exit_clean = True
+        except Exception as e:
             try:
-                output_queue.put_nowait(exc)
+                output_queue.put_nowait(e)
```
nicholaspun-wandb, line 5331: This can hang a `stream_logs()` consumer under slow-reader backpressure. The public `stream_logs()` path uses a bounded `output_queue`; if a terminal stream error arrives while that queue is full, this `put_nowait(e)` raises `QueueFull`, the exception is swallowed below, and `inner_exit_clean` stays false so no `None` sentinel is queued either. After the caller drains whatever log line filled the queue, the next `StreamReader` read waits forever instead of seeing the terminal error.

A unit repro is pretty small: construct a running sandbox, mock `StreamLogs` to return a single `LogEntry(error=LogStreamError(code="INVALID_RESUME_OFFSET", message="bad resume"))`, call `_stream_logs_async()` with `asyncio.Queue(maxsize=1)` prefilled with one string, then drain the prefilled item and assert the next `queue.get()` receives a `SandboxError`. On this branch that second get times out because the error was dropped. The read-file streaming path handles the same pattern by scheduling a guaranteed `output_queue.put(exc)` on `QueueFull`; `stream_logs()` should use the same guaranteed-delivery pattern for terminal errors.

djenriquez: Landed in 1439668: `stream_logs` now schedules `output_queue.put(exc)` on QueueFull instead of dropping the error, same as read_file streaming. Added `test_terminal_error_delivered_when_queue_full` for the prefilled queue case.

src/cwsandbox/_error_info.py
```diff
@@ -105,6 +105,32 @@
...
+CWSANDBOX_RUNNER_CAPACITY_EXHAUSTED = "CWSANDBOX_RUNNER_CAPACITY_EXHAUSTED"
+CWSANDBOX_PLACEMENT_REJECTED = "CWSANDBOX_PLACEMENT_REJECTED"
+CWSANDBOX_PLACEMENT_CONSTRAINT_UNSATISFIED = "CWSANDBOX_PLACEMENT_CONSTRAINT_UNSATISFIED"
+
+# Org / product gates that must not trigger placement spillover.
+CWSANDBOX_SERVERLESS_NOT_ALLOWED = "CWSANDBOX_SERVERLESS_NOT_ALLOWED"
+CWSANDBOX_SERVERLESS_DISABLED = "CWSANDBOX_SERVERLESS_DISABLED"
+CWSANDBOX_SERVERLESS_MISCONFIGURED = "CWSANDBOX_SERVERLESS_MISCONFIGURED"
+
+SPILLOVER_ELIGIBLE_REASONS: frozenset[str] = frozenset(
```
brandonrjacobs, line 126: Following up on the earlier spillover thread — understood the plan is a server-side reason fix with no SDK change. Two residual concerns: (1) until that fix is live, ordinary capacity exhaustion still surfaces as `CWSANDBOX_RUNNER_UNAVAILABLE` and spillover never fires — worth gating the 1.0.0 release on it (or handling `RUNNER_UNAVAILABLE` transitionally); (2) `CWSANDBOX_NO_SUITABLE_RUNNER`, `CWSANDBOX_RUNNER_OVERLOADED`, and `CWSANDBOX_SERVERLESS_CREATE_THROTTLED` are also outside `SPILLOVER_ELIGIBLE_REASONS` — if excluding them is deliberate, suggest pinning the eligible set in the spillover unit tests so it's contract rather than accident.

src/cwsandbox/_sandbox.py
```diff
@@ -1596,6 +1900,23 @@ def _from_sandbox_info(
...
+        sandbox._image_pull_credentials = None
+        sandbox._scratch_volume_names = (
+            tuple(
+                volume.name for volume in info._sandbox.spec.volumes if volume.HasField("scratch")
+            )
+            if isinstance(info, _SandboxView)
+            else ()
+        )
+        sandbox._service_urls = ()
+        sandbox._file_system_snapshot_ids = ()
```
brandonrjacobs, line 1990: `_from_sandbox_info` never initializes `_file_system_snapshot_id` (only the plural `_file_system_snapshot_ids`), and there's no `__getattr__` fallback — so any sandbox obtained via `Sandbox.from_id()` / `Sandbox.list()` raises `AttributeError` on the public `file_system_snapshot_id` property, and `stop(snapshot_on_stop=True)` hits the same `AttributeError` inside `_reject_unsatisfiable_snapshot_on_stop` instead of the documented `SnapshotOnStopConflictError`.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5033,130 @@ async def _stream_logs_async(
...
-                                # the inner loop with attempt_outcome
-                                # still "done"; the outer loop will set
-                                # done=True and the `finally` below cancels
-                                # the call and collector.  Any frames the
-                                # server pushed after `complete` would be
-                                # a protocol violation and are discarded.
                                 break
+                            if code == STREAM_TRUNCATED:
+                                raise SandboxStreamTruncatedError(msg)
+                            raise SandboxError(f"Log stream error: {msg}")
```
brandonrjacobs, line 5248: `STREAM_INTERRUPTED` isn't in `_STREAMING_FRESH_REINIT_CODES`, so it's treated as a terminal error and raises `SandboxError` — even though the session remains resumable server-side and the client is still holding `resume_log_session_id`/offset (the server's own message says to reconnect). This code is emitted during routine server-side restarts/deploys, so long-lived `follow=True` tails will die regularly. This one should resume rather than raise.

brandonrjacobs (commented):
Review at 35c8959 — 15 findings inline, ranked roughly by severity (correctness first). Happy to expand on any of them.

src/cwsandbox/_sandbox.py
```diff
@@ -3971,22 +4601,28 @@ async def _do_stop(
...
+                        )
+                    snapshot_volumes = list(self._scratch_volume_names)
+                request = sandbox_pb2.DeleteSandboxRequest(
                     sandbox_id=sandbox_id,
-                    graceful_shutdown_seconds=int(graceful_shutdown_seconds),
-                    file_system_snapshot_on_stop=snapshot_on_stop,
-                    max_timeout_seconds=max_timeout,
+                    grace_period_seconds=int(graceful_shutdown_seconds),
+                    snapshot_volumes=snapshot_volumes,
+                    allow_missing=missing_ok,
```
brandonrjacobs: This sends `allow_missing=True` together with `snapshot_volumes`, a combination the server rejects with `FAILED_PRECONDITION`/`CWSANDBOX_INVALID_REQUEST` when the sandbox is already gone — and the `missing_ok` catch below only suppresses `NOT_FOUND`. So `stop(snapshot_on_stop=True, missing_ok=True)` raises in exactly the missing case it documents suppressing (and because `allow_missing=True` is sent, the server never returns the `NOT_FOUND` that catch was written for).

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3489,513 @@ async def _start_async(self) -> str:
...
+                        getattr(translated, "reason", None) or e.code().name,
+                        self._placement_mode,
+                        alternate,
+                    )
+                    self._placement_mode = alternate
+                    if alternate == PlacementMode.SERVERLESS:
+                        self._runner_ids = None
+                    self._create_request_id = str(uuid.uuid4())
+                    continue
+                if primary_error is not None:
```
brandonrjacobs, line 3673: This restore runs on *any* attempt-2 `RpcError`, including ambiguous ones (`DEADLINE_EXCEEDED`, `UNAVAILABLE`) where the server may have already committed the spilled create under the spilled `request_id`. Restoring the original `_create_request_id` means a retry of `_start_async` (documented idempotent) re-runs attempt 1 with a request id the server never saw commit — potentially creating a second sandbox while the spilled one is running. Suggest restoring only on errors that guarantee attempt 2 didn't commit, and keeping the spilled request id on ambiguous failures.

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3489,513 @@ async def _start_async(self) -> str:
...
+            self._command is not None
+            or self._args is not None
+            or bool(self._environment_variables)
+            or "mounted_files" in explicit_keys
+            or bool({"volumes", "file_system_snapshot"} & explicit_keys)
+            or "secrets" in explicit_keys
+            or "resources" in explicit_keys
+        )
+        if container_field_overrides and self._container_image is None:
+            raise TypeError(
```
brandonrjacobs, line 3940: Every sparse override the docstring advertises (`command`, `*args`, `environment_variables`, `secrets`, `resources`, `mounted_files`, `volumes`, `file_system_snapshot`) raises this `TypeError` unless `container_image` is also passed — so `run_from_template('tmpl-1', command='python3')` and friends fail client-side before any RPC. Either the guard needs to relax or the signature/docstring shouldn't advertise per-field overrides.

src/cwsandbox/_sandbox.py
```diff
@@ -4270,20 +4932,21 @@ async def _snapshot_async(
...
             lambda: _create_snapshot_via_stub(
                 stub,
                 sandbox_id,
-                idempotency_key=effective_idempotency_key,
+                request_id=effective_request_id,
                 wait_for_ready=wait_for_ready,
                 auth_metadata=self._auth_metadata,
                 timeout=create_timeout,
-                max_timeout_seconds=create_max_timeout,
+                scratch_volume_name=(
```
brandonrjacobs: `snapshot()` hardwires `_scratch_volume_names[0]` with no way for the caller to choose. With both `volumes=` scratch volumes and a `file_system_snapshot` mount, the FSS volume lands after the `volumes=` entries, so `[0]` is not the FSS volume and this silently archives the wrong directory (the docstring promises the FSS-configured one). The API requires a name when multiple scratch volumes exist precisely to force disambiguation — suggest a `volume_name=` parameter, defaulting to the FSS volume or raising when ambiguous.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5033,130 @@ async def _stream_logs_async(
...
+                        if entry.log_session_id:
+                            session_id = entry.log_session_id
+                        if entry.next_log_offset:
+                            last_offset = int(entry.next_log_offset)
+                        chunk = entry.data.decode("utf-8", errors="replace")
+                        if not chunk:
+                            continue
+                        line_parts.append(chunk)
+                        line_parts_bytes += len(chunk.encode("utf-8", errors="replace"))
+                        if line_parts_bytes > MAX_LINE_BUFFER_BYTES:
```
brandonrjacobs: A single line over `MAX_LINE_BUFFER_BYTES` (1 MiB) now raises `SandboxStreamTruncatedError` and kills the whole stream — the previous client used the cap only to defer flushing and still delivered the line. One long minified/JSON log line shouldn't cost the consumer every subsequent line; consider flushing the oversized buffer as a line instead of raising. (The exception's docstring — "command ran to completion… output truncated in transit" — also doesn't describe this case: nothing completed and the abort is client-side.)

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5033,130 @@ async def _stream_logs_async(
...
-                                    pending_error = translated
-                                break
-                            if isinstance(item, Exception):
-                                pending_error = item
-                                break
+                request = sandbox_pb2.StreamLogsRequest(
+                    sandbox_id=sandbox_id,
+                    follow=follow,
+                )
+                if is_resume:
```
brandonrjacobs, line 5204: On an in-band fresh re-init (`SESSION_NOT_FOUND`/`REPLAY_GAP`/etc.) the rebuilt request carries only `sandbox_id` + `follow`, since these fields are set only when `attempt == 0`. Two consequences: (1) `follow=False` one-shot reads now silently restart from scratch and duplicate already-delivered output (the previous client raised for non-follow interrupts); (2) `timestamps=` silently flips off mid-stream. Formatting fields are only disallowed on resume requests, not fresh init, so they can be re-sent here — and non-follow reads should probably still raise rather than replay.

src/cwsandbox/_types.py
```diff
@@ -111,30 +111,158 @@ class ExecOutcome(StrEnum):
...
+        port: Container port the workload listens on.
+        name: Optional service name.
+        protocol: L4 protocol (defaults to TCP when unset).
+        visibility: Who may reach this port (PUBLIC/PRIVATE/CUSTOM).
+    """
+
+    port: int
+    name: str | None = None
+    protocol: ServiceProtocol | str | None = None
+    visibility: ServiceVisibility | str | None = None
```
brandonrjacobs, line 191: `visibility=None` goes on the wire as `VISIBILITY_UNSPECIFIED`, which the server treats as declared-but-never-exposed — so the most natural declaration, `Service(port=8080)`, silently yields no URL and `service_urls` stays empty forever, with no error or hint. Suggest defaulting to an explicit visibility, raising client-side when unset, or at minimum documenting that `visibility` is required for the service to be reachable.

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3489,513 @@ async def _start_async(self) -> str:
...
+                kwargs["gpu"] = g
+        return sandbox_pb2.Resources(**kwargs) if kwargs else None
+
+    def _apply_status_echo(self, view: _SandboxView) -> None:
+        """Refresh property echoes from a v1 Sandbox resource (create/Get/list)."""
+        self._scratch_volume_names = tuple(
+            volume.name for volume in view._sandbox.spec.volumes if volume.HasField("scratch")
+        )
+        status = view._sandbox.status
+        self._service_urls = tuple((s.port, s.name, s.url) for s in status.services if s.url)
```
brandonrjacobs: This only reads `ServiceStatus.url`; nothing in the SDK reads `ServiceStatus.endpoint.url`, which is where HTTPS endpoint-based services publish their URL (`url` stays empty for those). Since `service_urls` is the CHANGELOG's designated replacement for the removed v1beta2 address fields, those services are structurally invisible here. Suggest `s.url or s.endpoint.url`.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5033,130 @@ async def _stream_logs_async(
                     f"Sandbox {self._sandbox_id} is terminating"
                     " (follow=True requires a running sandbox)"
                 )
-
             if not self._is_done and not self._is_stopping:
```
brandonrjacobs, line 5175: The docstring and its example promise reading historical logs from a stopped sandbox with `follow=False`, but the v1 API rejects `StreamLogs` on terminated sandboxes (`FAILED_PRECONDITION`/`CWSANDBOX_SANDBOX_NOT_RUNNING`) — this gate only blocks `follow=True`, so the documented path sends the RPC and fails every time. Either reject client-side with a clear error or drop the promise from the docs.

src/cwsandbox/_sandbox.py
```diff
@@ -4038,6 +4692,14 @@ async def _do_stop(
                 self._stop_owned = True
                 sent_rpc = True
                 logger.info("Sandbox %s stop accepted, draining", sandbox_id)
+                if snapshot_ids and wait_for_ready:
+                    for snapshot_id in snapshot_ids:
+                        await _wait_for_snapshot_via_stub(
```
brandonrjacobs, line 4820: This archive wait runs without the `_retry_transient_rpc` guard used on the other snapshot surfaces, so a transient error during the (up to 600s/volume) wait fails `stop().result()` even though the delete was already accepted and the snapshot IDs captured. Worse, state is left `_Stopping` with `_stop_task` cleared, so retrying `stop(snapshot_on_stop=True)` raises `SnapshotOnStopConflictError` instead of re-joining the wait — the only recovery is manual `get_snapshot` polling. Suggest wrapping the poll in the transient-retry guard and letting a repeat `stop()` re-attach.

src/cwsandbox/_sandbox.py
```diff
@@ -5146,7 +5534,7 @@ async def request_generator() -> AsyncIterator[streaming_pb2.ExecStreamRequest]:
 
         # Queue decouples stream iteration from our processing.
         # Without this, processing suspends the stream and can cause issues.
-        response_queue: asyncio.Queue[streaming_pb2.ExecStreamResponse | Exception | None] = (
+        response_queue: asyncio.Queue[sandbox_pb2.ExecStreamResponse | Exception | None] = (
```
brandonrjacobs, line 5718: In the non-TTY collector fed by this queue, only `DEADLINE_EXCEEDED` gets translated; every other `AioRpcError` is enqueued untranslated and the consumer re-raises it as-is — so a transient `UNAVAILABLE` mid-exec surfaces raw `grpc.aio.AioRpcError` to callers following the documented `except SandboxError` contract. The TTY path routes the identical case through `_translate_rpc_error`; this path should too.

src/cwsandbox/_sandbox.py
```diff
@@ -3971,22 +4601,28 @@ async def _do_stop(
...
+                            "Cannot snapshot on stop: sandbox has no known scratch volumes",
+                            file_system_snapshot_id=None,
+                        )
+                    snapshot_volumes = list(self._scratch_volume_names)
+                request = sandbox_pb2.DeleteSandboxRequest(
                     sandbox_id=sandbox_id,
-                    graceful_shutdown_seconds=int(graceful_shutdown_seconds),
-                    file_system_snapshot_on_stop=snapshot_on_stop,
-                    max_timeout_seconds=max_timeout,
+                    grace_period_seconds=int(graceful_shutdown_seconds),
```
brandonrjacobs: `int()` truncates fractional grace, so `stop(graceful_shutdown_seconds=0.5)` sends `grace_period_seconds=0` — which v1 defines as immediate forced termination. A float-typed parameter invites fractions; suggest `math.ceil()` or rejecting sub-second values so a caller asking for half a second doesn't get an instant kill.

src/cwsandbox/_sandbox.py
djenriquez: Discovered sandboxes now initialize `_file_system_snapshot_id` in e5d5177. `file_system_snapshot_id` and `stop(snapshot_on_stop=True)` no longer raise AttributeError.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 resumes `follow=True` on STREAM_INTERRUPTED and keeps the session/offset. `follow=False` still raises so a one-shot read does not replay.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 does not send `allow_missing` with `snapshot_volumes`. A missing sandbox still returns NOT_FOUND, which `missing_ok` swallows.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 restores the original create id only when attempt 2 is a definite reject. Timeouts and unavailability keep the spilled id so a later `start()` cannot create a second sandbox.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 documents that `run_from_template` container-field overrides need `container_image`, and args need `command` plus the image.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 raises `SandboxSnapshotError` when `snapshot()` sees more than one scratch volume, instead of archiving `_scratch_volume_names[0]`.

src/cwsandbox/_sandbox.py
djenriquez: Left the raise in place so one oversized line ends the stream instead of dropping later lines. e5d5177 updates the `SandboxStreamTruncatedError` docstring to cover this client abort.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 re-sends `timestamps` on a fresh init and raises on `follow=False` instead of replaying. `tail_lines` and `since_time` stay first-attempt only.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 uses `ServiceStatus.endpoint.url` when `url` is empty, so HTTPS endpoint services show up in `service_urls`.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 removes the stopped-sandbox logs claim and example. v1 rejects StreamLogs on terminated sandboxes.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 runs non-timeout `AioRpcError`s on the non-TTY exec path through `_translate_rpc_error`, matching TTY.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 ceils fractional grace, so `0.5` becomes 1 second instead of a force-kill 0. Zero still means immediate stop.

src/cwsandbox/_error_info.py
djenriquez: Pinned `SPILLOVER_ELIGIBLE_REASONS` in the unit tests in e5d5177. Gating 1.0 on aviato#1739 and adding `RUNNER_UNAVAILABLE` to the eligible set are still open.

src/cwsandbox/_sandbox.py
djenriquez: e5d5177 retries transient Get errors inside the archive wait. A repeat `stop(snapshot_on_stop=True)` still conflicts; re-attach is not in this change.

brandonrjacobs (approved):

brandonrjacobs (changes requested):
Sorry, more findings found removing approval for now.

brandonrjacobs (commented):
Round 3 at c29cbb2c — 15 findings inline. Three are edge cases in the new review-fix commits (the spillover ambiguity classifier, the allow_missing handling, and the resume budget); the rest are new. Happy to expand on any of them.

src/cwsandbox/_sandbox.py
```diff
@@ -691,6 +818,98 @@ def _translate_rpc_error(
...
+    if isinstance(
+        exc,
+        (
+            SandboxUnavailableError,
+            SandboxTimeoutError,
+            SandboxResourceExhaustedError,
+        ),
+    ):
+        return False
+    return isinstance(exc, SandboxError)
```
brandonrjacobs: `_create_attempt_definitely_rejected` treats `INTERNAL`, `UNKNOWN`, `ABORTED`, and `CANCELLED` as definite rejections via the `isinstance(exc, SandboxError)` fallthrough — but none of those guarantee the create didn't commit (e.g. a server-side failure after the create was accepted surfaces as `INTERNAL`). On that shape the SDK restores the original `request_id`, and a retried `start()` can create a second sandbox while the spilled one's workload runs. Safer default: only classify as rejected on codes that assert the server refused the request (`INVALID_ARGUMENT`, `FAILED_PRECONDITION`, `PERMISSION_DENIED`, `NOT_FOUND`, `ALREADY_EXISTS`, typed `RESOURCE_EXHAUSTED`), and treat everything else as ambiguous — keeping the spilled id is replay-safe.

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3535,521 @@ async def _start_async(self) -> str:
...
+            if "memory_gb" in gpu:
+                g["memory_gb"] = int(gpu["memory_gb"])
+            if g:
+                kwargs["gpu"] = g
+        return sandbox_pb2.Resources(**kwargs) if kwargs else None
+
+    def _apply_status_echo(self, view: _SandboxView) -> None:
+        """Refresh property echoes from a v1 Sandbox resource (create/Get/list)."""
+        self._scratch_volume_names = tuple(
+            volume.name for volume in view._sandbox.spec.volumes if volume.HasField("scratch")
```
brandonrjacobs: This filters on `HasField("scratch")`, but per the v1 proto an UNSET `SandboxVolume.source` is also a valid scratch volume (scratch is the default when no source is given). A sandbox declaring `{name: "workspace"}` with no source and attached via `from_id()`/`list()` computes `_scratch_volume_names == ()`: `stop(snapshot_on_stop=True)` then raises "sandbox has no known scratch volumes" before any RPC, and a mix of explicit + default volumes silently archives only the explicit ones — quiet data loss. Suggest treating `source` unset as scratch, matching the server.

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3535,521 @@ async def _start_async(self) -> str:
...
+                "replace-on-presence container overrides require container_image; "
+                "the API replaces the whole container list and rejects sparse patches"
+            )
+        if self._container_image is not None:
+            if self._args is not None and self._command is None:
+                raise TypeError(
+                    "container overrides with args require command; "
+                    "the API rejects args without command after a whole-list container replace"
+                )
+            container = sandbox_pb2.PartialContainer(
```
brandonrjacobs, line 3950: `run_from_template(tid, container_image=...)` — the one override allowed without `command` — silently discards the template's entire container config: since the merge is replace-on-presence, sending `PartialContainer(name='main', image=...)` replaces the whole containers list, so the sandbox runs the bare image entrypoint with none of the template's command/args/env/mounted files/resources, and template volumes are left unmounted. The override also hardcodes the container name `main` and never sets `primary_container`, so templates whose container has another name end up pointing at a nonexistent container. At minimum the docstring should warn that `container_image` replaces (not patches) the template container; better, fetch-and-merge client-side or reject when the template container isn't named `main`.

src/cwsandbox/_sandbox.py
```diff
@@ -3953,40 +4643,45 @@ async def _do_stop(
...
                     sandbox_id=sandbox_id,
-                    graceful_shutdown_seconds=int(graceful_shutdown_seconds),
-                    file_system_snapshot_on_stop=snapshot_on_stop,
-                    max_timeout_seconds=max_timeout,
+                    grace_period_seconds=grace_seconds,
+                    snapshot_volumes=snapshot_volumes,
+                    # Snapshot-on-stop + allow_missing is rejected as
+                    # INVALID_REQUEST; omit the flag so a missing sandbox
+                    # still returns NOT_FOUND for the missing_ok catch.
+                    allow_missing=bool(missing_ok) and not snapshot_on_stop,
```
brandonrjacobs: Stripping `allow_missing` whenever `snapshot_on_stop` is set overcorrects: the combination is only rejected when the sandbox is absent — precisely so that a missing snapshot source can't silently succeed. With the flag stripped, a missing sandbox now returns `NOT_FOUND`, the `missing_ok` catch swallows it, and `stop(snapshot_on_stop=True, missing_ok=True)` returns clean success with `file_system_snapshot_id=None` — the caller asked for an archive and gets a success that masks there being none. Suggest passing `allow_missing` through and translating the server's rejection, or raising `SnapshotOnStopConflictError` client-side on the NOT_FOUND path.

src/cwsandbox/_error_info.py
brandonrjacobs, line 126: Even once the server-side capacity-taxonomy fix is live, two flagship scenarios still won't spill: an empty or no-capable CKS pool surfaces `CWSANDBOX_NO_SUITABLE_RUNNER`, and deferred placement surfaces `CWSANDBOX_RUNNER_OVERLOADED` — neither is in `SPILLOVER_ELIGIBLE_REASONS`. And on servers running today, ordinary exhaustion still surfaces as `CWSANDBOX_RUNNER_UNAVAILABLE`, so shipping 1.0.0 before that fix rolls out ships spillover as a documented no-op. Suggest adding `NO_SUITABLE_RUNNER` and `RUNNER_OVERLOADED` to the eligible set (and `RUNNER_UNAVAILABLE` transitionally, or gating the release on the server fix being live).

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3535,521 @@ async def _start_async(self) -> str:
...
-                raise _translate_rpc_error(e, operation="Start sandbox") from e
+                translated = _translate_rpc_error(e, operation="Create sandbox")
+                if (
+                    attempt == 1
+                    and primary_error is None
+                    and self._placement_spillover != PlacementSpillover.STRICT
+                    and _is_spillover_eligible(translated)
+                ):
+                    primary_error = translated
+                    alternate = (
```
brandonrjacobs, line 3652: After an ambiguous spill correctly keeps the flipped mode + spilled `request_id`, a later `start()` retry computes the spill target purely from the policy — there's no `alternate != current mode` guard. So the retry "spills" SERVERLESS→SERVERLESS and mints a fresh `request_id`, discarding the kept dedup token: if the earlier in-flight create committed, the new id creates a second sandbox — the exact collision keeping the spilled id was meant to prevent — and the retry burns its one hop on the same exhausted pool. Suggest skipping the spill when the alternate equals the current mode (and keeping the existing request id on same-mode retries).

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5097,141 @@ async def _stream_logs_async(
...
+                if is_resume:
+                    request.resume_log_session_id = session_id
+                    request.resume_log_offset = last_offset
+                else:
+                    # Fresh init (first attempt or after SESSION_NOT_FOUND /
+                    # REPLAY_GAP / runner loss). Re-send timestamps so
+                    # formatting survives a session reset; do not re-send
+                    # tail_lines / since_time — those replay a window.
+                    request.timestamps = timestamps
+                    if is_first_attempt:
```
brandonrjacobs: On a fresh re-init (`SESSION_NOT_FOUND`/`REPLAY_GAP`/`RUNNER_UNAVAILABLE`/`RUNNER_DRAINING` — routine during server deploys) the rebuilt request omits `tail_lines`/`since_time` entirely (they're gated on `attempt == 0`), so the server streams the entire log from the head — for a long-running sandbox that's hours of duplicate replay into a consumer that asked for `tail_lines=10`. Conversely, if attempt 0 dies before any frame, the retry drops the caller's window despite zero bytes delivered. The predicate should be "nothing delivered yet" rather than `attempt == 0`, plus `since_time≈now` (or a small tail) on true re-inits.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5097,141 @@ async def _stream_logs_async(
...
+                        if entry.log_session_id:
+                            session_id = entry.log_session_id
+                        if entry.next_log_offset:
+                            last_offset = int(entry.next_log_offset)
+                        chunk = entry.data.decode("utf-8", errors="replace")
+                        if not chunk:
+                            continue
+                        line_parts.append(chunk)
+                        line_parts_bytes += len(chunk.encode("utf-8", errors="replace"))
+                        if line_parts_bytes > MAX_LINE_BUFFER_BYTES:
```
brandonrjacobs: This checks the cumulative buffer against `MAX_LINE_BUFFER_BYTES` *before* splitting on newlines, so it raises even when the buffered data contains complete lines that would flush the buffer back under the limit — e.g. ~0.9 MiB of pending partial line plus one large multi-line frame (resume replays can arrive as a single large frame) dies at 1.1 MiB cumulative although no single line exceeds 1 MiB. It also makes 1 MiB of newline-free output (tqdm/pip `\r` progress bars) a fatal error, and contradicts the new docstring ("a single log line exceeded the client's 1 MiB buffer"). Suggest splitting and flushing complete lines first and capping only the retained partial line.

src/cwsandbox/_session.py
```diff
@@ -741,23 +759,30 @@ def compute(x: int, y: int) -> int:
...
                 temp_dir=temp_dir or self._defaults.temp_dir,
-                profile_ids=profile_ids,
-                profile_names=profile_names,
                 runner_ids=runner_ids,
                 resources=resources,
                 mounted_files=list(mounted_files) if mounted_files else None,
-                s3_mount=s3_mount,
-                ports=list(ports) if ports else None,
                 network=network,
+                services=list(services) if services else None,
```
brandonrjacobs: `list(services) if services else None` coerces an explicit empty override to `None`, so `@session.function(services=[])` falls back to the session defaults — re-exposing ports the caller explicitly cleared (reproduced at runtime: with `SandboxDefaults(services=(Service(port=8080, visibility=PUBLIC),))`, the function sandbox exposes 8080 PUBLIC despite `services=[]`). Same pattern for `volumes`. `session.sandbox([])` and `Sandbox(services=[])` honor the empty list, so behavior diverges by entry point. Should be `if services is not None`.

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5097,141 @@ async def _stream_logs_async(
...
-                            await queue.put(None)
-
-                    collect_task = asyncio.create_task(collect_responses(call, response_queue))
-
-                    # Outcome of this attempt: "done" (terminal), "resume"
-                    # (transient — try again), or "fresh" (server told us
-                    # the session is gone, drop resume state and re-init).
-                    attempt_outcome: str = "done"
-                    pending_error: Exception | None = None
+            while not done and attempt < STREAMING_RESUME_MAX_ATTEMPTS:
```
brandonrjacobs, line 5195: The reconnect budget (`STREAMING_RESUME_MAX_ATTEMPTS = 3`) is lifetime-scoped and never resets after a successful resume, so a long-lived `follow=True` stream permanently dies on the 3rd interruption — even if each resume succeeded and streamed for hours (`STREAM_INTERRUPTED` arrives on every server-side session teardown, so this is routine, not exceptional). The final raise also chains `from last_transport_error`, which at that point is `None` or an unrelated hours-old error. Suggest resetting the counter whenever a data frame is delivered — matching the poll-budget convention elsewhere in the SDK ("budget resets on any successful response").

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3535,521 @@ async def _start_async(self) -> str:
...
+
+    def _apply_status_echo(self, view: _SandboxView) -> None:
+        """Refresh property echoes from a v1 Sandbox resource (create/Get/list)."""
+        self._scratch_volume_names = tuple(
+            volume.name for volume in view._sandbox.spec.volumes if volume.HasField("scratch")
+        )
+        status = view._sandbox.status
+        service_urls: list[tuple[int, str, str]] = []
+        for s in status.services:
+            url = s.url or (s.endpoint.url if s.HasField("endpoint") else "")
```
brandonrjacobs, line 4030: `service_urls` is structurally empty for `custom`-visibility services: `ServiceStatus.url` is only assigned for PUBLIC/PRIVATE, and `endpoint.url` only for HTTPS endpoint services — so on custom-only fleets (which the e2e skip already acknowledges exist) no SDK surface exposes a connect address at all (`exposed_ports` carries no address either). Since the CHANGELOG points users migrating off `service_address` at `service_urls`, custom-visibility users have no migration path. Worth exposing whatever address surface custom services do get, or documenting the gap explicitly in the CHANGELOG migration note.

src/cwsandbox/_sandbox.py
```diff
@@ -1253,50 +1481,117 @@ def __init__(
...
-        )
+        self._max_lifetime_seconds = max_lifetime_seconds
+        if not from_template and self._max_lifetime_seconds is None:
+            self._max_lifetime_seconds = self._defaults.max_lifetime_seconds
 
-        self._tags: list[str] | None = self._defaults.merge_tags(tags)
-        self._environment_variables = self._defaults.merge_environment_variables(
-            environment_variables
+        self._tags: list[str] | None = (
+            list(tags or []) if from_template else self._defaults.merge_tags(tags)
```
brandonrjacobs: Template creates skip session-default merging (`list(tags or []) if from_template else self._defaults.merge_tags(tags)`), so `session.sandbox(template_id=...)` sandboxes carry no session tags — and `session.list()`, which filters by the merged default tags, can never find them: after a process crash they're un-adoptable and escape tag-scoped cleanup. The docstring promises "tags: Tags for the sandbox (merged with session defaults)" unconditionally. Same silent skip applies to session-default `environment_variables`/`annotations` on the template path. If the skip is intentional (template owns config), the docstring should say so; the tags half seems unintentional.

src/cwsandbox/_sandbox.py
```diff
@@ -4135,7 +4856,7 @@ async def _stop_async(
                         graceful_shutdown_seconds=graceful_shutdown_seconds,
                         missing_ok=missing_ok,
                         wait_for_ready=wait_for_ready,
-                        idempotency_key=idempotency_key,
+                        request_id=request_id,
                     )
                 )
                 self._stop_task.add_done_callback(self._on_stop_task_done)
```
brandonrjacobs, line 4926: When a second caller joins an already-in-flight stop task here, only the first caller's `missing_ok` reached `_do_stop` — the joiner's flag is silently ignored. So `stop()` followed by an external delete followed by `stop(missing_ok=True)` raises `SandboxNotFoundError` into the caller that asked to suppress it (the shield only catches `CancelledError`), and the reverse ordering gives a joiner silent success where it wanted the error. `snapshot_on_stop` got a join-conflict guard for exactly this coalescing problem; `missing_ok` needs the analogue (e.g. catch NOT_FOUND at the join when the joiner passed `missing_ok=True`).

src/cwsandbox/_sandbox.py
```diff
@@ -4394,375 +5097,141 @@ async def _stream_logs_async(
...
+                    # REPLAY_GAP / runner loss). Re-send timestamps so
+                    # formatting survives a session reset; do not re-send
+                    # tail_lines / since_time — those replay a window.
+                    request.timestamps = timestamps
+                    if is_first_attempt:
+                        if tail_lines is not None:
+                            request.tail_lines = tail_lines
+                        if since_time is not None:
+                            ts = timestamp_pb2.Timestamp()
+                            ts.FromDatetime(since_time)
```
brandonrjacobs: `since_time` is passed straight to `Timestamp.FromDatetime`, which treats naive datetimes as UTC — so `stream_logs(since_time=datetime.now() - timedelta(minutes=5))` from a UTC+2 caller asks for logs starting ~2h in the future (silently empty stream), and west-of-UTC callers get hours of extra replay. The docstring sets no tz requirement, and the SDK is asymmetric: inbound timestamps are normalized to UTC-aware, but outbound naive input is unguarded. Suggest rejecting naive datetimes or normalizing via `astimezone()`.

src/cwsandbox/_sandbox.py
```diff
@@ -3199,167 +3535,521 @@ async def _start_async(self) -> str:
...
+            if isinstance(placement, str):
+                placement = PlacementMode(placement.lower())
+            if placement == PlacementMode.SERVERLESS:
+                mode = sandbox_pb2.SANDBOX_MODE_SERVERLESS
+            elif placement == PlacementMode.CKS:
+                mode = sandbox_pb2.SANDBOX_MODE_CKS
+
+        runner_ids = self._runner_ids
+        if runner_ids and mode == sandbox_pb2.SANDBOX_MODE_SERVERLESS:
+            raise ValueError("runner_ids requires placement_mode=CKS")
```
brandonrjacobs, line 3879: `placement_spillover='serverless_then_cks'` combined with `runner_ids` is unusable: attempt 1 builds the SERVERLESS request while `runner_ids` is still set and raises `ValueError('runner_ids requires placement_mode=CKS')` before any RPC — and the message never mentions spillover. The CKS_THEN_SERVERLESS direction clears `runner_ids` for its serverless hop; the reverse direction should either hold the pins for the CKS attempt only (the natural "try serverless, fall back to my pinned runners" config) or reject the combination eagerly at construction with a spillover-aware message.

src/cwsandbox/_error_info.py
djenriquez: Added CWSANDBOX_NO_SUITABLE_RUNNER, CWSANDBOX_RUNNER_OVERLOADED, and CWSANDBOX_RUNNER_UNAVAILABLE to the spillover set. If the primary mode cannot place the request we try the alternate, and the primary reason is attached to a failed spill.

src/cwsandbox/_sandbox.py
djenriquez: The classifier is now an allowlist. INTERNAL/UNKNOWN/CANCELLED and other transport-shaped errors keep the spilled request id so a later start() cannot create a second sandbox.

src/cwsandbox/_sandbox.py
djenriquez: Unset volume source is treated as scratch, matching the proto default. from_id()/list() now see those volumes for snapshot-on-stop.

src/cwsandbox/_sandbox.py
djenriquez: Documented that container_image replaces the whole template container (command, args, env, files, resources). No fetch-and-merge.

src/cwsandbox/_sandbox.py
djenriquez: allow_missing is sent with snapshot_volumes. A CWSANDBOX_INVALID_REQUEST reject or a missing sandbox on snapshot-on-stop raises SnapshotOnStopConflictError instead of succeeding with no archive.

src/cwsandbox/_error_info.py
djenriquez: Those three reasons are now spill-eligible. Product gates, auth, and INVALID_ARGUMENT still do not spill.

src/cwsandbox/_sandbox.py
djenriquez: If the alternate mode already matches the current placement, we skip the hop and keep the existing request id.

src/cwsandbox/_sandbox.py
djenriquez: Fresh re-init resends tail_lines/since_time until any data arrives. After that, follow mode sets since_time to now and drops tail_lines so we do not replay hours of logs.

src/cwsandbox/_sandbox.py
djenriquez: Complete lines are flushed before the 1 MiB cap is applied. A single line over the cap still raises SandboxStreamTruncatedError.

src/cwsandbox/_session.py
djenriquez: session.function() now keeps empty services, volumes, and mounted_files instead of treating them as unset.

src/cwsandbox/_sandbox.py
djenriquez: The follow resume budget resets when a data frame is delivered. Exhaustion after in-band STREAM_INTERRUPTED no longer chains a None cause.

src/cwsandbox/_sandbox.py
djenriquez: Documented the gap: custom visibility often has no URL from the API, and the SDK does not invent one. Those services still appear in exposed_ports.

src/cwsandbox/_sandbox.py
djenriquez: Template creates now merge session tags and send them as a replace-on-presence override so list()/adopt can find the sandbox. Environment variables and annotations stay template-owned.

src/cwsandbox/_sandbox.py
djenriquez: A joiner that passed missing_ok=True now swallows SandboxNotFoundError from the shared stop task. The reverse (first caller swallowed, joiner wanted the error) is not recoverable.

src/cwsandbox/_sandbox.py
djenriquez: Naive since_time now raises ValueError. The timestamp must be timezone-aware.

src/cwsandbox/_sandbox.py
djenriquez: placement_spillover='serverless_then_cks' with runner_ids now raises at construction. Pins are not held across the serverless hop.

src/cwsandbox/_sandbox.py
djenriquez: The archive wait already retries transient Gets (c29cbb2). Re-attaching a later stop(snapshot_on_stop=True) onto a failed wait is a recovery-protocol change and is out of this pass.

brandonrjacobs (approved):

PR #156, opened 2026-08-21
feat: add DNS-name HTTPS egress grants on sandbox create

djenriquez:
## Summary

A sandbox can declare which hostnames it may reach over HTTPS when it is created. Exact names and a one-label wildcard (for example `*.pypi.org`) are accepted. A bare `*` is rejected because that is a fleet policy ceiling, not a per-sandbox grant. Turning off all egress cannot be combined with a hostname list. After create, the SDK reports the granted names from sandbox status.

DNS-name HTTPS egress at create

### Create-time hostname grants

The SDK sends hostname grants on sandbox create and reads them back from status.

- `src/cwsandbox/_types.py` — new `EgressRule` (`dns_name`); `NetworkOptions.egress` accepts rules or dicts, lowercases names, rejects empty/`*`, and rejects `deny_egress=True` with a non-empty list.
- `src/cwsandbox/_sandbox.py` — maps `egress` onto proto `egress[].dns_name`; `dns_egress_names` copies `status.effective_egress`.
- `src/cwsandbox/__init__.py` — exports `EgressRule`.
- `src/cwsandbox/_defaults.py`, `src/cwsandbox/_session.py` — docs; dict `network` still coerces through `NetworkOptions`.

### Tests

- `tests/unit/cwsandbox/test_types.py` — normalize, wildcard, `*` and empty rejection, dict coerce, deny-plus-names conflict.
- `tests/unit/cwsandbox/test_sandbox.py` — create mapping (typed and dict) and status echo.
- `tests/unit/cwsandbox/test_defaults.py` — nested egress dicts in `SandboxDefaults.from_dict`.
- `tests/integration/cwsandbox/test_dns_egress.py` — live grant vs miss against `pypi.org` / `example.com`; skips when the fleet cannot admit names.

### Docs

- `AGENTS.md`, `src/cwsandbox/AGENTS.md`, `tests/AGENTS.md` — public surface and file table.

## Behavior / Key decisions

| Input | Outcome |
| --- | --- |
| `pypi.org` or `*.pypi.org` | sent as `egress[].dns_name` on create |
| `"*"` or empty name | `ValueError` in the client |
| `deny_egress=True` and a non-empty hostname list | `ValueError` in the client |
| status `effective_egress` | `sandbox.dns_egress_names` |

This PR does not add CIDR, selector, or `any` destinations, `dns_name_except`, or a discovery `supports_dns_egress` flag. Vendored protos already had `dns_name`; they were not regenerated.

## Risk surface

This adds a public type (`EgressRule`) and a new field on `NetworkOptions`. Callers that omit `egress` keep deny-flag-only behavior. Hostname lists are create-only; the client does not send updates after create. The API-reference generator in `coreweave/docs` still needs `MANIFEST_GROUPS` updated for the new export (not in this repo).

## Test plan

- [x] `uv run pytest tests/unit` — 1295 passed in this worktree before commit
- [x] ruff format/check and mypy on `src/`
- [ ] `tests/integration/cwsandbox/test_dns_egress.py` — not run here (no `CWSANDBOX_API_KEY`)

### DIFF
--- a/src/cwsandbox/_sandbox.py
+++ b/src/cwsandbox/_sandbox.py
@@ -2939,8 +2939,8 @@ class Sandbox:
     def dns_egress_names(self) -> tuple[str, ...]:
         """Hostnames granted at create, echoed from ``status.effective_egress``.
 
-        Empty until a create/Get/list response reports name rules, when none
-        were requested, or after the sandbox stops.
+        Empty until a create/Get/list response reports name rules, or when
+        none were requested. Terminal sandboxes keep the last echoed names.
         """
         return self._dns_egress_names
 
--- a/src/cwsandbox/_sandbox.py
+++ b/src/cwsandbox/_sandbox.py
@@ -113,6 +113,7 @@ from cwsandbox._proto import (
 )
 from cwsandbox._resources import normalize_resources
 from cwsandbox._types import (
+    EgressRule,
     Endpoint,
     EndpointAuth,
     EndpointKind,
@@ -1363,6 +1364,7 @@ class Sandbox:
         runner_id: Runner ID where sandbox is running.
         returncode: Exit code if sandbox completed.
         started_at: When sandbox started running.
+        dns_egress_names: Hostnames granted at create (from status.effective_egress).
     """
 
     def __init__(
@@ -1428,8 +1430,9 @@ class Sandbox:
                 use ``sandbox.write_file()`` after the sandbox is running.
             s3_mount: Removed in 1.x; passing a value raises ``TypeError``.
             ports: Removed in 1.x; use ``services=[Service(...)]`` instead.
-            network: Deny-flag ``NetworkOptions`` (or dict). Port exposure uses
-                ``services=``.
+            network: ``NetworkOptions`` (or dict) with deny flags and optional
+                create-time hostname grants (``egress=[EgressRule(dns_name=...)]``).
+                Port exposure uses ``services=``.
             placement_mode: ``PlacementMode`` or string (``serverless`` / ``cks``).
             placement_spillover: ``PlacementSpillover`` or string. Default
                 ``strict`` (no create retry). Non-strict modes retry CreateSandbox
@@ -1531,6 +1534,7 @@ class Sandbox:
         self._image_pull_credentials: ImagePullCredentials | None = None
         self._scratch_volume_names: tuple[str, ...] = ()
         self._service_urls: tuple[tuple[int, str, str], ...] = ()
+        self._dns_egress_names: tuple[str, ...] = ()
         self._file_system_snapshot_ids: tuple[str, ...] = ()
         # Use explicit resources or fall back to defaults, then normalize
         effective_resources = (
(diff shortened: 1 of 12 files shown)


brandonrjacobs (commented):
Nice, focused SDK surface for create-time hostname grants. A few small gaps below — nothing blocking on correctness of the happy path.

src/cwsandbox/_types.py
```diff
@@ -250,19 +250,77 @@ def __post_init__(self) -> None:
...
+
+    def __post_init__(self) -> None:
+        name = self.dns_name.strip().lower()
+        if not name:
+            raise ValueError("EgressRule.dns_name cannot be empty")
+        if name == "*":
+            raise ValueError(
+                'EgressRule.dns_name cannot be "*"; that is a policy ceiling, not a sandbox grant'
+            )
+        object.__setattr__(self, "dns_name", name)
```
brandonrjacobs, line 295: **P2 — grammar vs docs.** Docs say exact or a single leftmost wildcard (`*.pypi.org`), but `__post_init__` only rejects empty/`*`. Values like `*.*.pypi.org`, `*pypi.org`, `foo.*.com`, or `pypi.org:443` are accepted here and fail only at create. Please enforce the documented grammar (and add unit rejects).

src/cwsandbox/_sandbox.py
```diff
@@ -2929,6 +2935,15 @@ def service_urls(self) -> tuple[tuple[int, str, str], ...]:
         """
         return self._service_urls
 
+    @property
+    def dns_egress_names(self) -> tuple[str, ...]:
+        """Hostnames granted at create, echoed from ``status.effective_egress``.
+
+        Empty until a create/Get/list response reports name rules, when none
+        were requested, or after the sandbox stops.
```
brandonrjacobs: **P3 — docstring.** Says empty “after the sandbox stops,” but stop does not clear `_dns_egress_names`, and terminal state stops refreshing status echoes. Align the text with caching behavior (or clear on stop if emptiness is intended).

tests/integration/cwsandbox/test_dns_egress.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def _skip_if_dns_egress_unavailable(exc: BaseException) -> None:
+    """Skip when the fleet cannot admit a dns_name create."""
+    if isinstance(exc, SandboxValidationError):
+        fields = " ".join(v.field for v in exc.field_violations)
+        if "dns_name" in f"{fields} {exc}".lower():
+            pytest.skip(f"runner policy does not admit DNS-name egress: {exc}")
+    if isinstance(exc, SandboxError) and exc.reason == CWSANDBOX_NO_SUITABLE_RUNNER:
+        pytest.skip(f"no runner advertises DNS-name egress: {exc}")
```
brandonrjacobs: **P3 — skip coverage.** Also skip `CWSANDBOX_PLACEMENT_CONSTRAINT_UNSATISFIED` (and similar unenforceable/placement reasons) so environments that cannot host name grants soft-skip instead of hard-failing.

src/cwsandbox/_types.py
djenriquez: Addressed. `EgressRule` now rejects names that are not a DNS-1123 subdomain or a single leftmost wildcard, using the same grammar create uses.

src/cwsandbox/_sandbox.py
djenriquez: Addressed. The docstring no longer says names go empty on stop; terminal sandboxes keep the last echoed names.

tests/integration/cwsandbox/test_dns_egress.py
djenriquez: Addressed. The test also skips `CWSANDBOX_PLACEMENT_CONSTRAINT_UNSATISFIED`. Other spillover reasons still fail the test.

brandonrjacobs (approved):

PR #158, opened 2026-08-25
fix(list): accept include_stopped as alias for show_terminated

joseph-shih-wandb:
## Summary
- `wandb==0.28.2` CLI still calls `Sandbox.list(include_stopped=...)` after the v1 rename to `show_terminated`, which makes `wandb beta sandbox ls` crash (`SBX-CLI-002`).
- Accept `include_stopped` as an alias on `Sandbox.list` and `Session.list` so the released W&B CLI keeps working until wandb 0.28.3 (already fixed in wandb/wandb#12482).

## Context
v1 `ListSandboxesRequest` uses `show_terminated` (field 5):

sandbox.proto
Ln 970–979
```
message ListSandboxesRequest {
  ...
  bool show_terminated = 5 [(google.api.field_behavior) = OPTIONAL];
```

v1beta2 has the old name: `include_stopped = 6` on `ListSandboxesRequest` in proto/coreweave/sandbox/v1beta2/gateway.proto.

## Test plan
- [x] `uv run pytest tests/unit/cwsandbox/test_sandbox.py tests/unit/cwsandbox/test_session.py tests/unit/cwsandbox/test_defaults.py -k "include_stopped or ResolveShowTerminated or show_terminated"` → 9 passed
- [x] `Sandbox.list(include_stopped=False)` (wandb 0.28.2 default `ls` path) does not TypeError
- [x] `Sandbox.list(include_stopped=True)` sets proto `show_terminated=True` (unit)
- [x] Live CLI: `wandb==0.28.2` + this branch via `PYTHONPATH`, `wandb beta sandbox ls --entity rnd-team-sandboxes-enabled --output json` → rc=0, parseable JSON
- [ ] After this ships on PyPI, nightly SBX-CLI-002 should pass on `wandb==0.28.2` without waiting for a wandb release

Made with [Cursor](https://cursor.com)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,7 +89,7 @@ Advanced configuration kwargs (for `run()`, `run_from_template()`, `Session.sand
 
 Class methods:
 - `Sandbox.session(defaults)`: Create a `Session` for managing multiple sandboxes (sync)
-- `Sandbox.list(tags=None, status=None, runner_ids=None, show_terminated=False, ...)`: Query existing sandboxes, return `OperationRef[list[Sandbox]]`. Use `.result()` to block or `await` in async contexts. By default, terminal sandboxes (completed, failed, terminated) are excluded. Set `show_terminated=True` to include them. `profile_ids` / `profile_names` raise `TypeError`.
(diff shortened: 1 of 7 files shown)


joseph-shih-wandb:
Close it since nightly installs wandb>=0.28.2 (today 0.28.2) and cwsandbox>=1.1.1 (today 1.7.0). 

The crash is the 0.28.2 CLI still passing include_stopped. wandb already renamed that in https://github.com/wandb/wandb/pull/12482. 

When 0.28.3 (or whatever next) is on PyPI, nightly woukld pick it up and ls works with current cwsandbox

PR #159, opened 2026-08-28
feat: add multi-container support

djenriquez:
## Summary

A sandbox can run more than one container in the same pod. One container is the primary and owns the sandbox lifetime. Extra containers are helpers that share localhost and optional volumes. Exec, file read and write, logs, and an interactive shell can name a helper; omitting the name still hits the primary, so existing one-container calls keep working.

Multi-container sandbox

### Create and types

Callers pass a list of containers on create and on session sandbox construction. More than one container requires exactly one primary, a name on every row, and resources on every row. GPU is allowed only on the primary. The list form cannot be mixed with the single-container image, command, args, resources, mounted files, secrets, pull credentials, or environment kwargs.

Volumes stay at sandbox scope. Sharing is two containers mounting the same volume name. A convenience mount path on a volume or snapshot option attaches to the primary only.

The existing command-and-args create path still sends one container named `main` and does not set the primary flag.

- `src/cwsandbox/_types.py` — container and volume-mount types, name rules, create validation
- `src/cwsandbox/_sandbox.py` — create, plus echo of the container list and per-container status
- `src/cwsandbox/_defaults.py` — shareable container list on defaults
- `src/cwsandbox/_session.py` — session sandbox accepts the list
- `src/cwsandbox/__init__.py` — public exports

### Targeting

Exec, read/write file, stream logs, and shell take an optional container name. Empty or omitted targets the primary.

- `src/cwsandbox/_sandbox.py` — container field on those RPCs
- CLI `exec`, `files`, `logs`, `sh` — `--container`
- CLI `get` — prints the container list and status

### API stubs

Vendored v1 stubs include the primary flag on each container.

### Docs

`AGENTS.md` documents the list form, targeting, and optional mount paths.

## Behavior / Key decisions

| Input | Outcome |
|---|---|
| Create with command and args (one container) | One container named `main`; primary flag unset |
| Create with a container list, more than one row | Exactly one primary; every row named and resourced |
| GPU on a non-primary row | Rejected at the client |
| Container name omitted on exec, files, logs, or shell | Primary |
| Container name set | That container |
| Volume with a mount path | Mounted on the primary |
| Volume without a mount path | Declared only; attach via each container's mounts |
| Container list plus image, command, or resources kwargs | `TypeError` |
| Create from a template with a container list | Replaces the template's container list |
| Remote function decorator | Still one container |

## Risk surface

One-container create, exec, files, and logs keep the same call shape. The public API reference generator needs the new exported types added to its manifest. A helper that never starts keeps the sandbox from reaching running. After a helper has started, a crash there does not fail the sandbox.

## Test plan

- [x] `uv run pytest tests/unit/cwsandbox/test_multi_container.py` plus the CLI, defaults, snapshot, sandbox, and utilities modules touched here — 642 passed
- [x] `uv run pytest tests/integration/cwsandbox/test_multi_container.py` against staging — 3 passed: named exec and files isolation, shared volume, Python primary reaching Redis on loopback

Not covered live: logs and shell targeting, CLI `--container`, templates, and session defaults.

### DIFF
--- a/src/cwsandbox/_sandbox.py
+++ b/src/cwsandbox/_sandbox.py
@@ -181,6 +181,7 @@ from cwsandbox._types import (
     _coerce_container,
     _coerce_object_storage_access,
     _coerce_volume_mount,
+    _unique_secrets_by_env_var,
     _validate_containers,
 )
 from cwsandbox.exceptions import (
@@ -334,13 +335,14 @@ class ContainerStatus:
     Attributes:
         name: Container name.
         state: Observed lifecycle state of this container.
-        exit_code: Exit code once the container is COMPLETED or FAILED.
+        exit_code: Exit code once the container is in a terminal state.
+            None while the container is still running.
         restart_count: Times the container has restarted.
     """
 
     name: str
     state: SandboxStatus
-    exit_code: int = 0
+    exit_code: int | None = None
     restart_count: int = 0
 
 
@@ -2003,20 +2005,7 @@ class Sandbox:
             Secret(**s) if isinstance(s, dict) else s for s in (secrets or ())
         ]
         if merged_secrets:
-            seen: dict[str, Secret] = {}
-            for secret in merged_secrets:
-                env_var = secret.env_var
-                assert env_var is not None  # guaranteed by Secret.__post_init__
-                if env_var in seen and secret != seen[env_var]:
-                    raise ValueError(
-                        f"Conflicting secrets for env_var {env_var!r}: "
-                        f"Secret(store={seen[env_var].store!r}, name={seen[env_var].name!r}, "
-                        f"field={seen[env_var].field!r}) vs "
-                        f"Secret(store={secret.store!r}, name={secret.name!r}, "
-                        f"field={secret.field!r})"
-                    )
-                seen[env_var] = secret
-            self._start_kwargs["secrets"] = list(seen.values())
+            self._start_kwargs["secrets"] = list(_unique_secrets_by_env_var(merged_secrets))
 
         self._channel: grpc.aio.Channel | None = None
         self._stub: sandbox_pb2_grpc.SandboxServiceStub | None = None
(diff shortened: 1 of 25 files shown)


djenriquez:
Stacked on #164 (`feat/runtime-class-security-subpath`). This branch is rebased onto that PR so one-container create-spec fields land first, and multi-container stays the follow-up.

scott-cw (commented):
Two findings from a code review of the branch diff.

src/cwsandbox/_sandbox.py
```diff
@@ -4514,6 +4749,87 @@ def _apply_status_echo(self, view: _SandboxView) -> None:
             self._resource_limits = self._resources_from_proto(status.effective_resources)
             self._resource_requests = self._resource_limits
             self._resource_gpu = self._gpu_from_proto(status.effective_resources)
+        echoed = tuple(self._container_from_proto(row) for row in view._sandbox.spec.containers)
```
scott-cw, line 4759: High (correctness): Echoed containers are rebuilt through `Container.__post_init__`, applying the create-path input validator (reserved names, DNS-1123, `working_dir != "/"`) to server-owned data. One non-conforming container anywhere in the org raises a bare `ValueError` out of `start()`, `get_status()`, `from_id()`, and all of `Sandbox.list()`. Not a `CWSandboxError`, so it's fatal in `_classify_poll_error`. The egress/ingress echo above already guards this.

src/cwsandbox/_sandbox.py
```diff
@@ -1681,18 +1763,55 @@ def __init__(
         self._auth = auth if auth is not None else self._defaults.auth
 
         from_template = template_id is not None
+        effective_containers = (
+            containers
+            if containers is not None
+            else (None if from_template else self._defaults.containers)
+        )
+        using_containers = effective_containers is not None
```
scott-cw, line 1774: Medium (correctness): When `containers` comes from `SandboxDefaults` rather than kwargs, inherited `environment_variables`, `resources`, `security_context`, `working_dir`, and `secrets` are silently zeroed. The identical conflict via kwargs raises `TypeError` at 1794. A `Session` with org-wide secret injection loses it the moment one `session.sandbox()` call passes `containers`.

src/cwsandbox/_sandbox.py
djenriquez: Echoed containers now skip the create-time name and working_dir checks, so a reserved sidecar or working_dir="/" no longer fails list/get/poll. Creating a Container(...) still rejects those.

src/cwsandbox/_sandbox.py
djenriquez: Leaving this. containers= replaces those single-container defaults instead of inheriting them, because there is no safe automatic home for session secrets on a sidecar list. Docs now say to put secrets, env, and working_dir on each Container.

brandonrjacobs (approved):

PR #160, opened 2026-08-31
feat(auth): add per-client auth strategies, enable W&B auth

nicholaspun-wandb:
### Description

Builds out auth strategies in the client and enables using W&B auth patterns natively within the client. Users can select `AuthStrategy.WANDB` to use the credential resolution patterns from [W&B SDK](https://github.com/wandb/wandb).

### Changes

  - Add AuthStrategy.COREWEAVE_API_KEY and AuthStrategy.WANDB
  - Add per-Sandbox and per-Session auth configuration
  - Preserve CoreWeave Bearer authentication as the default
  - Resolve W&B credentials lazily through the optional W&B dependency
  - Send W&B credentials using x-wandb-api-key
  - Preserve the legacy set_auth_mode() hook for compatibility
  - Support custom AuthProvider implementations

### Testing

Added unit tests and validated that a sandbox can be started using my W&B credentials: 
```
from cwsandbox import AuthStrategy, Sandbox

with Sandbox.run(
    auth=AuthStrategy.WANDB,
    max_lifetime_seconds=240,
    container_image="python:3.11",
) as sb:
    sb.wait(timeout=60)
    print(f"Sandbox started: {sb.sandbox_id}", flush=True)
    print(f"Runner (tower) ID: {sb.runner_id}", flush=True)

    echo_result = sb.exec(["echo", "hello"]).result()
    print(echo_result.stdout, flush=True)

    file_content = b"Hello, World!\n"
    sb.write_file("/tmp/data.txt", file_content).result()
    print("write_file: '/tmp/data.txt'", flush=True)

    read_back = sb.read_file("/tmp/data.txt").result()
    decoded = read_back.decode("utf-8", errors="replace").rstrip()
    print(f"read_file bytes={len(read_back)} content={decoded}", flush=True)

    cat_result = sb.exec(["cat", "/tmp/data.txt"]).result()
    print(f"cat /tmp/data.txt -> {cat_result.stdout.rstrip()}", flush=True)
```


src/cwsandbox/_auth.py
```diff
@@ -37,81 +38,186 @@ def __bool__(self) -> bool:
...
+        raise CWSandboxAuthenticationError(
+            "No W&B credentials found. Set WANDB_API_KEY or run `wandb login`."
+        )
+
+    if not isinstance(auth, wbauth.AuthApiKey):
+        raise CWSandboxAuthenticationError(
+            "CWSandbox currently supports only W&B user API-key authentication."
+        )
+
+    headers = {
```
djenriquez, line 123: **High: Forward the W&B host**

`_resolve_wandb_auth()` selects credentials for `settings.base_url`, but drops that host when it builds the gRPC metadata. With `WANDB_BASE_URL=https://acme.wandb.io`, the request has `x-wandb-api-key` but no `x-wandb-host`. The backend uses `x-wandb-host` to route dedicated W&B auth; without it, the request falls through to the SaaS authenticator and the key is rejected.

Please include a normalized bare `host[:port]` for non-SaaS hosts, omit it for `api.wandb.ai`, and add coverage for both cases.

src/cwsandbox/__init__.py
```diff
@@ -286,7 +286,10 @@ def _to_awaitable(w: Waitable) -> asyncio.Task[Any]:
 
 __all__ = [
     "__version__",
+    "AuthConfig",
```
djenriquez, line 291: **Medium: Register the new public exports**

`AuthConfig`, `AuthStrategy`, and `AuthProvider` are now in `cwsandbox.__all__`, but the docs manifest still only covers `AuthHeaders` and `set_auth_mode`. The manifest-drift test fails on all three names, so reference generation stops.

Please add them to `MANIFEST_GROUPS`. `AuthConfig` is a type alias, so it also needs alias rendering or an explicit exclusion. The new public `auth=` parameters should be added to the affected `Args:` sections as well.

djenriquez (commented):
This is close, but the dedicated W&B host path needs one fix before I can approve. I also left a docs note for the new public exports.

src/cwsandbox/_auth.py
nicholaspun-wandb: Fixed in ad08e4a. The resolver now parses the W&B SDK resolved credential host once, strips userinfo, lowercases the hostname, validates and preserves only non-default explicit ports, and classifies SaaS using the normalized hostname. Coverage includes api.wandb.ai, api.qa.wandb.ai, uppercase SaaS, SaaS with :443, dedicated hosts with userinfo, and dedicated default/non-default ports. The full client suite passes (1,325 tests).

src/cwsandbox/__init__.py
nicholaspun-wandb: Addressed in both repositories. ee27ba8 documents auth on every affected public Args section in cwsandbox-client. coreweave/docs#3651 registers AuthStrategy and AuthProvider in MANIFEST_GROUPS, explicitly excludes the AuthConfig type alias, and adds generator coverage.

djenriquez (approved):

PR #161, opened 2026-08-31
docs(examples): add run_from_template example

joseph-shih-wandb:
Adds the missing template-consumption example to `examples/`: inherit an org template as-is via `Sandbox.run_from_template()`, then override the container to demonstrate the replace-on-presence contract (container overrides require `container_image`; `args` requires `command`).

Also adds index rows to `examples/README.md` and `examples/AGENTS.md`.

Part of the sandbox examples ask (templates/policy coverage); the JS SDK already ships the equivalent `run-from-template.ts`.

## Verification

- Ruff format/lint pass.
- **Run live end-to-end** against prod (org template created with `cwic sandbox template create`): block 1 inherited the template and exec'd successfully, block 2 replaced the container and exec'd successfully, both sandboxes stopped cleanly on context exit. Throwaway template deleted afterward.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/AGENTS.md
+++ b/examples/AGENTS.md
@@ -21,6 +21,7 @@ Runnable example scripts demonstrating cwsandbox SDK usage patterns.
 | `stdin_streaming.py` | Sync/Async | `def main()` | Stdin streaming with `exec(stdin=True)` |
 | `function_decorator.py` | Sync | `def main()` | Remote function execution with `@session.function()` |
 | `resource_configuration.py` | Sync | `def main()` | ResourceOptions, flat dict, nested dict, GPU, response properties |
+| `run_from_template.py` | Sync | `def main()` | `Sandbox.run_from_template()`: inherit a template, replace-on-presence container override |
 | `error_handling.py` | Sync | `def main()` | Exception hierarchy: SandboxExecutionError, TimeoutError, NotFoundError |
 | `file_system_snapshots.py` | Sync | `def main()` | FSS snapshot/restore/fork with `file_system_snapshot`, `snapshot()`, snapshot management |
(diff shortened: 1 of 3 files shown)


joseph-shih-wandb:
Applied external-review finding: tags supplied via `defaults` are sent as a **replace-on-presence override** — the created sandbox carries them *instead of* the template's tags, not merged with them (`overrides.tags.extend(self._tags)` in `_build_create_from_template_request`). The example's comment, docstring, and README bullet previously said "merged"; now corrected, with a note that dropping the tags keeps the template's.

joseph-shih-wandb:
Applied both remaining findings:

- Dropped the "inherit unchanged / as-is" phrasing (docstring, section heading, README, AGENTS row) — the block is now described as "start from a template with no spec overrides", which is accurate given tags are still sent as an override.
- The no-override block no longer execs into the template's arbitrary image (a distroless image would have no `sh`); it waits for RUNNING and prints status instead. The exec demo lives only in the override block, where the image is pinned by construction.

Re-ran live end-to-end: block 1 reaches `running`, block 2 execs in the replaced container, no sandboxes left behind.

joseph-shih-wandb:
Applied: the first block is now a true zero-override create — `Sandbox.run_from_template(template_id)` and nothing else, so the sandbox keeps the template's own tags. The tag-override demonstration (with the replace-not-merge caveat) moved to the container-override block, which was already an override showcase. Re-ran live: block 1 reaches `running` carrying the template's tags, block 2 execs in the replaced container, nothing left behind.

djenriquez (approved):

PR #162, opened 2026-08-31
feat: add direct sandbox data connections

brandonrjacobs:
## Summary

- vendor the public `ConnectSandbox` and `SandboxDataPlaneService` protobuf surface
- add `DataPlaneMode.AUTO`, `GATEWAY`, and `DIRECT`
- generate operation-scoped P-256 credentials in memory for direct mTLS exec, logs, and file operations
- keep lifecycle operations and bearer credentials on the API gateway
- cap the process-wide idle pool at 64 channels and detach unavailable or retiring channels immediately while existing streams drain
- retry only structured pre-admission shard-retirement rejections; ambiguous `UNAVAILABLE` operations are never replayed
- split caller-provided upload chunks into frame-safe messages

## Compatibility and rollout

`AUTO` remains the default. It gives direct setup a one-second total budget, then falls back to the gateway and observes a 30-second cooldown. `DIRECT` retains a ten-second readiness window and never falls back. `GATEWAY` is the deterministic rollback setting.

## SDK usage

```python
from cwsandbox import DataPlaneMode, Sandbox, SandboxDefaults

# AUTO is the default: prefer direct mTLS and fall back to the gateway.
with Sandbox.run() as sandbox:
    result = sandbox.exec(["echo", "hello"]).result()
    sandbox.write_file("/tmp/result", result.stdout.encode()).result()

# Require the direct path when validating connectivity.
with Sandbox.run(data_plane_mode=DataPlaneMode.DIRECT) as sandbox:
    print(sandbox.read_file("/etc/hostname").result().decode())

# Force the gateway path for deterministic rollback.
defaults = SandboxDefaults(data_plane_mode=DataPlaneMode.GATEWAY)
with Sandbox.run(defaults=defaults) as sandbox:
    print(sandbox.exec(["echo", "gateway"]).result().stdout)
```

Sandbox creation, status, and lifecycle calls continue through the API. Only exec, logs, and file operations use the selected data path.

## Scale and security

Private keys remain in the client process. Direct RPCs carry no bearer token, and each certificate requests only the operation it authorizes. Active streams drain on their existing channel; once that channel is marked stale, new calls create a replacement. Inactive direct channels use the bounded shared cache.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src/`
- `pytest` — 1,335 passed
- `pre-commit run --all-files`
- opt-in strict direct integration coverage for exec, logs, write, and read

Coordinated API-reference manifest update: coreweave/docs#3659.

The strict live smoke still needs to be rerun in an environment with direct endpoint issuance enabled.


brandonrjacobs:
Follow-up: `stream_logs` now cancels and releases each direct attempt before resume, so follow reconnects cannot retain idle mTLS leases.

djenriquez (commented):
I found one blocking correctness issue in the channel reuse path and one small API documentation follow-up. The rest of the transport and security changes look ready.

src/cwsandbox/_direct.py
```diff
@@ -0,0 +1,405 @@
...
+        self._lock = asyncio.Lock()
+
+    async def acquire(
+        self,
+        bundle: _CredentialBundle,
+        *,
+        timeout: float,
+    ) -> _DirectChannelLease:
+        async with self._lock:
+            entry = self._entries.get(bundle.cache_key)
```
djenriquez, line 109: **High: drop connections to retiring shards**

`AUTO` keeps this cached channel after a direct RPC returns `UNAVAILABLE/CWSANDBOX_RUNNER_SHARD_RETIRING`. The runner keeps the HTTP/2 connection healthy while rejecting new RPCs, so later operations reuse the same channel instead of falling back. A focused repro made two direct `ReadFile` calls, never called the gateway, and left the pool entry cached.

Please discard the channel for this exact pre-admission rejection and retry once on a fresh direct connection, or use the gateway in `AUTO`. Rechecking readiness alone will not help because the connection remains `READY`.

src/cwsandbox/__init__.py
```diff
@@ -291,6 +292,7 @@ def _to_awaitable(w: Waitable) -> asyncio.Task[Any]:
     "CWSandboxAuthenticationError",
     "CWSandboxError",
     "CWSandboxValidationError",
+    "DataPlaneMode",
```
djenriquez, line 296: **Medium: add `DataPlaneMode` to the API manifest**

`DataPlaneMode` is now public, so the repo's release process also requires adding it to `MANIFEST_GROUPS` in `coreweave/docs/scripts/cwsandbox-api-ref/generate.py`. Please include or link the coordinated docs change so the generated API reference does not omit it.

src/cwsandbox/_direct.py
brandonrjacobs: Addressed in 930f556. Any direct `UNAVAILABLE` now discards the pooled channel, and the structured retiring-shard rejection retries once on a fresh lease. The regression test verifies the first lease is discarded and not reused.

src/cwsandbox/__init__.py
brandonrjacobs: Added the coordinated API-manifest update in coreweave/docs#3659 and linked it from this PR description.

djenriquez (approved):

PR #163, opened 2026-09-01
feat(sandbox): add HTTPS endpoint request timeout

matthoare117-wandb:
## Summary

- HTTPS `Endpoint` accepts optional `request_timeout_seconds`. Omit/`0` keeps the platform default (15s on serverless). The SDK only checks that the value is an `int`; Aviato currently accepts `0` or `[15, 900]`.
- This is the server-side product HTTPS clock (504 while the sandbox stays alive). It is not `Sandbox.run(request_timeout_seconds=...)`, which stays the client RPC deadline.
- Create and create-from-template copy a nonzero value onto proto `EndpointSpec`. Vendored v1 stubs are refreshed from aviato (`SandboxSpec.primary_container` is reserved; the SDK still sends one container and omits `Container.primary`).

```python
Service(
    port=8080,
    name="http-timeout",
    visibility=ServiceVisibility.PUBLIC,
    endpoint=Endpoint(
        kind=EndpointKind.HTTPS,
        auth=EndpointAuth.OPEN,
        request_timeout_seconds=120,
    ),
)
```

`Sandbox` still only exposes `service_urls`. Applied timeout is not echoed.

## Test plan

- [x] `uv run pytest tests/unit` (1316 passed in this worktree)
- [ ] Live staging/seabed create with `request_timeout_seconds=120`, then `GET /ready` and a ~2 min `/within` (skip if `CWSANDBOX_HTTPS_REQUEST_TIMEOUTS_NOT_SUPPORTED`)


brandonrjacobs:
## Review nits

Non-blocking; feature LGTM otherwise.

### 1. Prefer user-facing naming (please change)

In `_types.py`, `CHANGELOG.md`, and `AGENTS.md`, replace “Aviato” with user-facing wording (“the API” / “the server” / “CoreWeave Sandbox”) when describing the accepted timeout range.

### 2. Template override + `0`

Replace-on-presence keeps the template scalar when the override is zero, so `request_timeout_seconds=0` on a template create cannot reset a template’s `120` back to the 15s default. Docs currently read as omit/`0` ⇒ platform default; that holds for plain create, not for clearing a template value. Worth one clarifying sentence.

### 3. Merge with #162

Both refresh `sandbox_pb2*` and touch `_sandbox.py` (`primary_container`). Expect a stub rebase. #162 also sets `Container.primary=True` while this omits it — either is fine for a single container today; pick one after merge.

### 4. Commit history

Five commits include add-then-remove of a client-side minimum. Squash or tidy before release if you care about public bisectability.

### 5. Optional follow-ups

- Applied timeout is on `EndpointStatus` but not exposed on `Sandbox` (already noted in the PR).
- No README example (CHANGELOG/AGENTS may be enough).

brandonrjacobs (approved):

PR #169, opened 2026-09-03
feat(sandbox): add TLS passthrough product endpoints

matthoare117-wandb:
## Summary

- Create-time `Endpoint(kind=TLS_PASSTHROUGH)` on a PUBLIC service. Auth and `request_timeout_seconds` must be omitted (`None`, not `0`).
- Create, Get, list, and `from_id` fill `Sandbox.service_addresses` as `TlsPassthroughEndpointStatus` (`port`, `name`, `kind`, `address` as `host:port`). Use the host as TLS SNI; the workload owns certs.
- TLS addresses stay off `service_urls`. On a live handle, a later CREATING/RUNNING Get keeps a cached address per service `(port, name)` when that row is still present and Get omits the endpoint or address. `from_id` / list have no Create cache. Terminal states clear it.

```python
Service(
    port=443,
    name="tls",
    visibility=ServiceVisibility.PUBLIC,
    endpoint=Endpoint(kind=EndpointKind.TLS_PASSTHROUGH),
)
```

See `examples/tls_passthrough.py`.

## Test plan

- [x] `uv run pytest tests/unit/cwsandbox/test_types.py tests/unit/cwsandbox/test_sandbox.py` (538 passed)
- [x] Live create + `from_id` + list + SNI GET via `examples/tls_passthrough.py` on api-stg (`product-tls-ok`)


src/cwsandbox/_sandbox.py
```diff
@@ -4695,27 +4715,57 @@ def _apply_status_echo(self, view: _SandboxView) -> None:
...
+                        )
                     )
-                )
+            elif has_endpoint and kind == sandbox_pb2.ENDPOINT_KIND_TLS_PASSTHROUGH:
+                tls_rows.append((s.port, s.name, s.endpoint.address))
+            elif not has_endpoint and s.url:
+                service_urls.append((s.port, s.name, s.url))
         self._service_urls = tuple(service_urls)
         self._service_endpoints = tuple(service_endpoints)
+        if status.state in (
```
nicholaspun-wandb, line 4744: Could we clear cached TLS addresses for every state other than `CREATING` and `RUNNING`? The backend suppresses `EndpointStatus.address` outside those two live states, but this denylist lets `PAUSED` retain the cached address. It also lets a poll-source `STATE_UNSPECIFIED` retain it because `_apply_status_echo()` runs before that state is mapped to `COMPLETED`. I reproduced both cases: the resulting status is `PAUSED`/`COMPLETED` while `service_addresses` still contains the old target. An allowlist for `STATE_CREATING` and `STATE_RUNNING`, plus regression tests for `PAUSED` and poll-source `UNSPECIFIED`, would preserve the server contract.

brandonrjacobs, line 4744: Please clear cached TLS addresses unless `status.state` is `STATE_CREATING` or `STATE_RUNNING`.

The Gateway only echoes `EndpointStatus.address` in those two states. A denylist for terminal/`TERMINATING` still retains addresses after `PAUSED`, and after poll-source `STATE_UNSPECIFIED` (mapped to `COMPLETED` only after `_apply_status_echo`). Prefer an allowlist for the two live states, plus regression tests for `PAUSED` and poll `UNSPECIFIED`.

brandonrjacobs (changes requested):
One change needed before merge: TLS address caching should clear outside live states.

nicholaspun-wandb (dismissed):

nicholaspun-wandb (approved):

brandonrjacobs (approved):

PR #170, opened 2026-09-03
feat: add Sandbox.run_from_file for Compose create

djenriquez:
## Summary

The Python SDK can create a sandbox from a pull-only Compose YAML file. Callers pass a filesystem path or raw bytes, name the primary service, and may set image overrides or default CPU and memory. The file bytes are sent unchanged, so whitespace is part of request identity. After the API accepts the create, waiting for the sandbox to run is the same poll as a normal create: early states are walked through, and preparing is not treated as ready. Image builds are out of scope; leftover build stanzas come back as not implemented.

A live integration test starts Redis, a Python API that waits on Redis, and an Ubuntu primary from one Compose file, then checks they can talk over loopback once the sandbox is running.

Create sandbox from Compose file

## What changed

Callers get a new factory next to ordinary create and create-from-template. It reads Compose bytes, applies local guards, and starts through the create-from-file RPC.

From-file inherits sandbox-level defaults (tags, network, placement mode, runner pin, annotations, object-storage access, max lifetime) and ignores container, volume, service, resource, snapshot, secret, and runtime-class defaults. Placement spillover must be strict.

Vendored stubs include the new RPC and a preparing status. Wait still polls through preparing. Docs, changelog, and an example script are updated.

## Details

| Input | Outcome |
| --- | --- |
| path `str` / `Path` | Opened as a file; file bytes go on the wire |
| `bytes` | Sent as-is |
| Compose text passed as `str` | Treated as a path (encode to bytes instead) |
| empty `primary_service` | `ValueError` |
| contents over 256 KiB | `ValueError` |
| GPU on `default_resources` | `ValueError` (the API also rejects this) |
| volumes, services, `runtime_class`, `image_pull_credentials`, `network_ids`, `build_contexts`, other ordinary-create-only kwargs | `TypeError` |
| non-strict `placement_spillover` | `TypeError` (a second from-file with a different mode is a different hash) |
| leftover Compose `build:` | `CWSANDBOX_NOT_IMPLEMENTED` |
| same `request_id`, different bytes | `CWSANDBOX_REQUEST_ID_CONFLICT` |
| Get after create | Translated spec only; no source-file echo |

Deliberate omissions: no client Compose validator, no gzip-tar helper, no session factory from-file, and no mapping of proto startup probes onto the public container type.

New public file-type export; the API reference generator manifest is not updated here. Ordinary create and create-from-template are unchanged except the start branch and the named preparing status. Callers that still treat unknown proto statuses as unspecified would have mapped preparing incorrectly; this change names it and keeps wait polling through it.

## Test plan

- [x] `uv run pytest tests/unit/cwsandbox/test_sandbox.py -k "from_file or preparing"` — 12 passed
- [x] `uv run pytest tests/unit` — 1466 passed
- [x] `uv run mypy src/cwsandbox/_sandbox.py src/cwsandbox/_types.py src/cwsandbox/__init__.py` — clean
- [x] `uv run pytest tests/integration/cwsandbox/test_run_from_file.py` against the staging API — `test_run_from_file_dependent_services` passed


nicholaspun-wandb (approved):

joseph-shih-wandb (approved):

src/cwsandbox/_sandbox.py
```diff
@@ -1094,13 +1099,18 @@ def _resolve_placement_for_spillover(
...
 
     For non-``STRICT`` spillover, an unset/unspecified primary is filled in as
-    the spillover's first mode (CKS or serverless). Templates only allow
-    ``STRICT``.
+    the spillover's first mode (CKS or serverless). Templates and from-file
+    creates only allow ``STRICT``.
     """
+    if from_file and spillover != PlacementSpillover.STRICT:
+        raise TypeError(
+            f"placement_spillover must be 'strict' for run_from_file (got {spillover.value!r})"
```
brandonrjacobs, line 1112: Nit: from-file raises `TypeError` here while templates raise `ValueError` two lines below for the same constraint. Prefer matching the template exception type (or documenting the intentional difference).

src/cwsandbox/_sandbox.py
```diff
@@ -2321,6 +2405,126 @@ def run_from_template(
...
+            image_overrides: Per-service pullable image refs. Keys must be
+                services in ``contents``.
+            default_resources: CPU/memory copied onto each service that omitted
+                ``deploy.resources``. Per container, not a project budget.
+                GPU is rejected locally (Gateway also rejects it).
+            defaults: Optional ``SandboxDefaults``. Tags, network,
+                ``placement_mode``, ``runner_ids``, annotations, object-storage
+                access, and max lifetime are inherited. Container/volume/
+                service defaults are not. Non-strict ``placement_spillover``
+                on defaults is ignored (from-file is always strict).
```
brandonrjacobs, line 2463: Nit: true for defaults (forced `STRICT`), but an explicit `placement_spillover=cks_then_serverless` raises. Worth one phrase so callers don’t assume silent ignore for the kwarg too.

tests/unit/cwsandbox/test_sandbox.py
```diff
@@ -6672,6 +6841,16 @@ def test_from_proto_terminating(self) -> None:
...
 
+    def test_from_proto_preparing(self) -> None:
+        from cwsandbox._proto import sandbox_pb2
+        from cwsandbox._sandbox import _RUNNING_STATUSES, _TERMINAL_STATUSES
+
+        status = SandboxStatus.from_proto(sandbox_pb2.STATE_PREPARING)
+        assert status == SandboxStatus.PREPARING
+        assert status.to_proto() == sandbox_pb2.STATE_PREPARING
+        assert SandboxStatus.PREPARING not in _TERMINAL_STATUSES
+        assert SandboxStatus.PREPARING not in _RUNNING_STATUSES
```
brandonrjacobs, line 6852: Suggestion: this locks enum membership, but not that `_poll_until_stable` treats `PREPARING` as transient. A short PREPARING→RUNNING poll mock would pin the contract that motivated naming this status (otherwise proto `STATE_PREPARING` used to map to `UNSPECIFIED` and stop `wait()` early).

src/cwsandbox/_sandbox.py
```diff
@@ -288,6 +290,8 @@ class SandboxStatus(StrEnum):
     Attributes:
         PENDING: Sandbox has been accepted but not yet scheduled.
         CREATING: Sandbox container is being created.
+        PREPARING: Accepted image fill; not placed. Gateway does not insert
```
brandonrjacobs, line 293: Nit: the class docstring summary above still says `CREATING -> RUNNING` and doesn’t mention `PREPARING`.

brandonrjacobs (approved):
Approve — solid pull-only `CreateSandboxFromFile` surface. Request shape, defaults inheritance, strict spillover, and the `PREPARING` naming fix all look right; CI is green.

Left a few non-blocking nits inline. Follow-up (already called out in the PR): API reference generator manifest, plus `network_ids` / `build_contexts` when Bob builds land.

Also nit (not on a diff line): the `SandboxStatus` summary still says `CREATING -> RUNNING` while attributes document `PREPARING`.

src/cwsandbox/_sandbox.py
djenriquez: Agreed — this is a bad value, not a bad keyword. #173 raises ValueError so it matches template creates.

src/cwsandbox/_sandbox.py
djenriquez: Agreed. The defaults bullet now says an explicit non-strict keyword still raises. In #173.

tests/unit/cwsandbox/test_sandbox.py
djenriquez: Agreed — enum membership did not pin the poll loop. #173 adds a wait() mock that goes PREPARING then RUNNING.

src/cwsandbox/_sandbox.py
djenriquez: Agreed. The lifecycle line now names PREPARING. In #173.

